PR #4, opened 2024-03-23
Create python-package.yml

jpodivin:


.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest
```
TomasTomecek, line 30: I'm assuming pytest is there for the future test runs, right?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
jkonecny12, line 37: I would rather see this as local configuration. I mean create pep8 configuration file as
https://pep8.readthedocs.io/en/release-1.7.x/intro.html#configuration

and pylint configuration file and having Makefile or other easy way to start checkers locally without GH actions. GH actions would then just executed Makefile command.

If you want I can do PR for that.

jpodivin:
Implemented as part of a different PR.

PR #8, opened 2024-04-02
Renaming the project

jpodivin:


FrostyX:
+1 for the name change but we should probably decide whether `logdetective` or `log-detective` and be consistent about it.

TomasTomecek:
> +1 for the name change but we should probably decide whether `logdetective` or `log-detective` and be consistent about it.

Agreed! So far we have mostly marketed the project as "Log Detective" so I think we can still decide. The website is available under both names.

xsuchy (approved):
+1 to PR
+1 to "logdetective" as the name of cli tool.

jpodivin:
I think that we should have the same name as is the CLI invocation. And since `logdetective` is lot easier to write than `log-detective` I'm voting for that.

TomasTomecek (approved):
okay, let's keep logdetective then :)

PR #10, opened 2024-04-07
Chunks instead of lines for drain

jpodivin:
The original log will be split into chunks. At this time the chunks will be delimited by whitespace or backslash. This does cover most of the cases I have encountered so far.

A more general solution would be list of patterns. However those would require more significant testing.


TomasTomecek (approved):
I wonder if this could be a bottleneck in the future on huge texts once we run accelerated on a GPU

PR #13, opened 2024-04-12
Fix README file

jkonecny12:
There was a typo and requirements.txt is not used in this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ To analyze a log file, run the script with the following command line arguments:
 
 Example usage:
 
-    logdetective https://example.com/logs.txt --model https://mymodel.co/model_file --summarizer drain --nlines 10
+    logdetective https://example.com/logs.txt --model https://mymodel.co/model_file --summarizer drain --n_lines 10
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,7 @@ To analyze a log file, run the script with the following command line arguments:
 
 Example usage:
 
-    logdetective https://example.com/logs.txt --model https://mymodel.co/model_file --summarizer drain --nlines 10
```
jpodivin, line 22: That's actually correct.

jpodivin (changes requested):

README.md
jkonecny12: Really? It didn't worked for me and based on the source code it should be as `--n_lines`:

https://github.com/fedora-copr/logdetective/blob/master/logdetective/logdetective.py#L176

I'm getting:

```
logdetective: error: unrecognized arguments: --nlines 10
```

jpodivin: Sorry I misread the diff.

jpodivin (approved):

PR #15, opened 2024-04-17
user friendly logging - INFO is default level, add -q and progressbar

xsuchy:
Previously the logdetective does not print anything for long time and seems to be stuck. I changed default level to INFO.
I added code to have meaningfull use of -v and not require -vvvvvvvvvv to get to INFO level. And I added progressbar when downloading the model. For integration I added --quite that surpress logging and only print result.


pyproject.toml
```diff
@@ -13,11 +13,11 @@ requests = "^2.31.0"
 llama-cpp-python = "^0.2.56"
 tiktoken = "^0.6.0"
 drain3 = "^0.9.11"
+progress = "*"
```
jpodivin, line 16: Version should be pinned to a sensible default.

jpodivin:
Couple of things:

Adding `progress` into dependencies, should be ideally through `poetry add progress`, this also pins the version to sensible default and updates lockfile for faster installation. 

Second, I'm not sure that the library you are using actually has the module you import. The https://github.com/verigak/progress/tree/master documentation indicates that the import should be `progress.bar`, it also doesn't provide `ProgressBar` class. 

Maybe you wanted to use a different lib?

jpodivin (changes requested):
Couple of things:

Adding `progress` into dependencies, should be ideally through `poetry add progress`, this also pins the version to sensible default and updates lockfile for faster installation. 

Second, I'm not sure that the library you are using actually has the module you import. The https://github.com/verigak/progress/tree/master documentation indicates that the import should be `progress.bar`, it also doesn't provide `ProgressBar` class. 

Maybe you wanted to use a different lib?

xsuchy:
I used python3-progress-1.6-11.fc40.noarch and it works with that.

```
$ python
Python 3.12.2 (main, Feb 21 2024, 00:00:00) [GCC 14.0.1 20240217 (Red Hat 14.0.1-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import progressbar
>>> progressbar.ProgressBar
<class 'progressbar.bar.ProgressBar'>
```

xsuchy:
> Adding progress into dependencies, should be ideally through poetry add progress, this also pins the version to sensible default and updates lockfile for faster installation.

Updated.

pyproject.toml
xsuchy: This is a sensible default.
Poetry suggested I use' 1.6`, But I disagree with that. It works with 1.5, too (and any lower version), And it will likely continue to work with 2.0, too. IMO, there is no need to restrict it; it would just add an additional chore in the future.

jpodivin:
> progressbar.ProgressBar

It doesn't work on 39 with `python3-progress-1.6-9.fc39`.
And neither does it work with the one installed from PyPi. 
I suppose there must be some downstream patches between 1.6-9 and 1.6-11. I don't really know why would there be.
But it does show that we can't rely on the API being stable between versions :/ 

```
Python 3.12.2 (main, Feb 21 2024, 00:00:00) [GCC 13.2.1 20231205 (Red Hat 13.2.1-6)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import progressbar
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'progressbar'
>>> 
```

xsuchy:
My fault:
```
$ python
Python 3.12.2 (main, Feb 21 2024, 00:00:00) [GCC 14.0.1 20240217 (Red Hat 14.0.1-0)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import progressbar
>>> progressbar.__file__
'/usr/lib/python3.12/site-packages/progressbar/__init__.py'
>>> 
$ rpm -qf /usr/lib/python3.12/site-packages/progressbar/__init__.py
python3-progressbar2-3.53.2-11.fc40.noarch
```

jpodivin (approved):

PR #22, opened 2024-05-07
Enable pypi release

jkonecny12:
This PR contains a few parts:
* Add script to automate version bumping of the project
* Prepare pyproject.toml for releases on Pypi

TODO (future PR): Ideally Pypi release should be automated when new version is detected

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,26 @@
 [tool.poetry]
 name = "logdetective"
 version = "0.1.0"
-description = ""
+description = "Log using LLM AI to search for build/test failures and provide ideas for fixing these."
 authors = ["Jiri Podivin <jpodivin@gmail.com>"]
 license = "Apache-2.0"
 readme = "README.md"
 include = ["logdetective/drain3.ini"]
+classifiers = [
+    "Development Status :: 4 - Beta",
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: Apache Software License",
+    "Natural Language :: English",
+    "Programming Language :: Python :: 3",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+    "Topic :: Software Development :: Debuggers",
+    "Topic :: Internet :: Log Analysis"
+]
+
+[tool.poetry.urls]
+homepage = "https://github.com/fedora-copr/logdetective"
+issues = "https://github.com/fedora-copr/logdetective/issues"
 
 [tool.poetry.dependencies]
 python = "^3.11"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,8 @@ logdetective = 'logdetective.logdetective:main'
 
 [tool.pylint]
 disable = [
+    "inconsistent-return-statements",
     "missing-module-docstring",
-    "too-few-public-methods"
+    "too-few-public-methods",
+    "unspecified-encoding",
 ]
--- a/scripts/bumpver.py
+++ b/scripts/bumpver.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+import argparse
+import subprocess
+
+
+NAME = "logdetective"
+
+
+def call_poetry(bump_rule):
+    """Call poetry tool to bump the version."""
+    # Bump the version based on the bump rule
+    # See poetry version --help
+    ret = subprocess.run(["poetry", "version", "-s", bump_rule], capture_output=True, check=True)
+
+    return ret.stdout.decode().strip()
+
+
+def bump_version(target_version=None):
+    """Bump pyproject.toml version."""
+    if target_version:
+        version = call_poetry(target_version)
+    else:
+        version = call_poetry("patch")
+
+    print(f"Version bumped to {version}")
+
+    return f"{version}"
+
+
+def bump_patch_version(version):
+    """Add +1 to the current version."""
+    major, minor, patch = map(int, version.strip().strip('"').split("."))
+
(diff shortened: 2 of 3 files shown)


scripts/bumpver.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/python3
```
jpodivin, line 1: I would recommend:
`#!/bin/env python` instead to avoid issues with virtual environments.

It shouldn't really make much of difference with the current state of the script (no external libs), but in case things change.

pyproject.toml
```diff
@@ -1,11 +1,26 @@
...
+    "License :: OSI Approved :: Apache Software License",
+    "Natural Language :: English",
+    "Programming Language :: Python :: 3",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+    "Topic :: Software Development :: Debuggers",
+    "Topic :: Internet :: Log Analysis"
+]
+
+[tool.poetry.urls]
+Homepage = "https://github.com/fedora-copr/logdetective"
```
jpodivin, line 22: I believe these should be in lower case. https://python-poetry.org/docs/pyproject/#urls

scripts/bumpver.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if target_version is None:
+        target_version = bump_patch_version(version.strip().strip('"'))
+
+    # Update version in the lines
+    lines[version_line_index] = f'{version_prefix}= "{target_version}"\n'
+
+    # Write back to the file
+    with open(filename, 'w') as f:
+        f.writelines(lines)
```
jpodivin, line 38: I think you can avoid this logic entirely by using https://docs.python.org/3/library/tomllib.html Since we are on the 3.11 the library is now in builtins.

jpodivin (changes requested):
Handful of things I would like to see changed. Also, I'm wondering if you are going to implement automation for pushing tags to PyPi?

scripts/bumpver.py
jkonecny12: That was my first version but additional library didn't really simplify the code, so I went without it.

However, I missed that,there is new buildin solution.

jkonecny12:
Yes, Pypi automatic releasing will be added after this. I split these to not block this anymore.

scripts/bumpver.py
jkonecny12: Fixed.

pyproject.toml
jkonecny12: I have it from some example but I don't think it's correct so - Fixed.

scripts/bumpver.py
jkonecny12: I tried your recommended library and nice improvement. Thanks for recommendation.

jkonecny12:
**Updated**

scripts/bumpver.py
xsuchy: I, on the other hand, prefer `#!/usr/bin/python3` because `#!/bin/env python` says, "Run it with any python interpreter you have specified in your environment.  And that is not what you test and intend to do.

xsuchy (approved):
LGTM now

jpodivin (changes requested):
I've just remembered something https://python-poetry.org/docs/cli/#version . This way we can effectively avoid dealing with toml completely. It will also give us sanity checks for free. The same invocation of poetry version will give us new version of the package for tag/patch. Maybe you could wrap the poetry command, pass the rule to poetry trough our script and embed the rule used in commit message.

jkonecny12:
**UPDATED**

scripts/bumpver.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+if __name__ == "__main__":
+    # Parse arguments
+    parse = argparse.ArgumentParser("bumpver.py",
+                                    description="Bump version of the logdetective project.")
+    parse.add_argument("-t", "--tag", action="store_true",
+                       dest="tag", help="Create also git a tag.")
+    parse.add_argument("--version", action="store", default=None,
+                       dest="version", help="Create also git a tag.")
```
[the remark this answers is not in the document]
jkonecny12: Fixed.

scripts/bumpver.py
```diff
@@ -0,0 +1,102 @@
...
+                       help="Disable checks on the code and lock file before commit.")
+    args = parse.parse_args()
+
+    if args.check:
+        if not run_checks():
+            print("\nChecks returned error!", file=sys.stderr)
+            print("If you want to bypass checks use '--no-check' argument.")
+            sys.exit(1)
+
+    # Replace 'pyproject.toml' with your actual file path
```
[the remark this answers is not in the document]
jkonecny12: Yes, fixed.

scripts/bumpver.py
```diff
@@ -0,0 +1,104 @@
...
+def bump_version(target_version=None):
+    """Bump pyproject.toml version."""
+    if target_version:
+        version = call_poetry(target_version)
+    else:
+        version = call_poetry("patch")
+
+    print(f"Version bumped to {version}")
+
+    return f"{version}"
```
jpodivin, line 30: Small issue. The returned string is `packagename version`. This breaks things down the line if you want to run tag.

jpodivin (changes requested):

scripts/bumpver.py
jkonecny12: Good point! I even found there is `-s` option to remove all the string even from the first call so making even more optimal.

jpodivin (approved):

PR #29, opened 2024-05-17
readme: add real example

TomasTomecek:
Since we don't have a demo yet, let's have a real example in our readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,70 @@ Or if the log file is stored locally:
     logdetective ./data/logs.txt
 
 
+Real Example
+------------
+Let's have a look at a real world example. Log Detective can work with any logs though we optimize it for build logs.
+
+We're going to analyze a failed build of a python-based library that happened in Fedora Koji buildsystem:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,6 +42,33 @@ Or if the log file is stored locally:
...
+```
```
xsuchy, line 67: The lines are really loong. What about wrapping them? https://stackoverflow.com/questions/41238148/how-to-wrap-long-lines-inside-of-markdown-code-in-github-and-gitlab-issu

TomasTomecek:
we can strip those mega long lines, indeed they are confusing

README.md
TomasTomecek: thanks for the suggestion, will wrap

TomasTomecek:
is this more readable?

the reason I sent this PR is that 2 people last week asked me about a demo - I think showing a real example is a good start

xsuchy (approved):
Much better for me. 
+1 from me.

jpodivin (approved):

PR #30, opened 2024-05-29
Added example of how to use a different model to readme

Fan:
I was a bit confused when trying to run logdetective with different models as I did not realize that the download URL had to be passed into the --model argument. I added a bit of text to explain this as well as provided a quick example.


TomasTomecek (approved):
sweet! thanks

PR #31, opened 2024-06-03
remove dependency on tiktoken

xsuchy:
The strict dependency on tiktoken version is causing a problem: https://bugzilla.redhat.com/show_bug.cgi?id=2284439
I look how we are using it, but I could not find any use of tiktoken in our code.
It was added by @jpodivin in the initial commit. Is it possible that it is a leftover? Can it be removed?


jpodivin (approved):
You are correct, this is indeed a leftover from previous iteration of the project.

PR #32, opened 2024-06-07
Download models using HF hub API

jpodivin:
Adds dependency on huggingface_hub.
Removes builtin download function.
Simplifies model initialization.
Additional comments provided to improve code literacy. Cache management now handled by huggingface_hub.
Removed dependency on progressbar2.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ python = "^3.11"
 requests = "^2.31.0"
 llama-cpp-python = "^0.2.56"
 drain3 = "^0.9.11"
-progressbar2 = ">=3.53.2"
+huggingface-hub = "^0.23.2"
 
 [build-system]
 requires = ["poetry-core"]
--- a/logdetective/constants.py
+++ b/logdetective/constants.py
@@ -1,6 +1,6 @@
 
 # pylint: disable=line-too-long
-DEFAULT_ADVISOR = "https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF/resolve/main/mistral-7b-instruct-v0.2.Q4_K_S.gguf?download=true"
+DEFAULT_ADVISOR = "TheBloke/Mistral-7B-Instruct-v0.2-GGUF"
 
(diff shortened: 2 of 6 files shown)


jkonecny12 (approved):
Looks good to me. Thanks

TomasTomecek (approved):
I love the code simplification. Worked locally just fine even though I had to download the model, again, since huggingface hub fetches the bare git repo.

One thing we should discuss at some point is about the relation of the CLI and server parts: can they share some internals? meaning, none of these changes affect the server part.

LGTM, well done

jpodivin:
> I love the code simplification. Worked locally just fine even though I had to download the model, again, since huggingface hub fetches the bare git repo.
> 
> One thing we should discuss at some point is about the relation of the CLI and server parts: can they share some internals? meaning, none of these changes affect the server part.
> 
> LGTM, well done

Thanks. I would actually prefer to package models with the image directly, instead of downloading them during startup. It would be easier to ensure the right model is used and remove potential security hazard. We can include hash checks in the layer as well, to make sure we are getting what we asked for.

pyproject.toml
```diff
@@ -30,7 +30,7 @@ python = "^3.11"
 requests = "^2.31.0"
 llama-cpp-python = "^0.2.56"
 drain3 = "^0.9.11"
-progressbar2 = ">=3.53.2"
+huggingface-hub = "^0.23.2"
```
xsuchy, line 33: This is not packaged for Fedora :(

TomasTomecek: whoops :/ we forgot about this

the package itself looks okay, but the extras are really demanding

https://github.com/huggingface/huggingface_hub/blob/main/setup.py

xsuchy: Do we need extras?

jpodivin: No we don't. HF treats this as "almost" a metapackage, different types for different use cases. For our purposes we can stick with minimal subset of functionality and things will work. I'll confirm this and get back to you though, just to be sure.

PR #34, opened 2024-06-24
Action for publishing on tag release

jpodivin:
Documentation for the new release process is included.

### DIFF
--- a/PUBLISHING
+++ b/PUBLISHING
@@ -1,0 +1,26 @@
+# Publishing new release
+
+New releases must be published on PyPi, ideally through github action,
+although manual release management is also permissible.
+In all cases, make sure to follow rules of [semantic versioning](https://semver.org).
+
+New releases should be created in two stages.
+
+## Release patch workflow:
+
+1. Create new branch with name indicating it contains new release, e.g. `release-v5.7.1`.
+2. Increment version in pyproject.toml using `poetry version` command. Use appropriate rule.
+3. Commit the change in pyproject.toml.
+4. Push the change, open PR and wait for approval.
+
+## Github workflow:
+
+1. Open releases dialog in repo main page.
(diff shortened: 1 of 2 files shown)


PUBLISHING
```diff
@@ -0,0 +1,25 @@
...
+
+New releases must be published on PyPi, ideally through github action,
+although manual release management is also permissible.
+In all cases, make sure to follow rules of [semantic versioning](https://semver.org).
+
+New releases should be created in two stages.
+
+## Release patch workflow:
+
+1. Create new branch with name indicating it contains new release.
```
TomasTomecek, line 11:
```suggestion
1. Create new branch with name indicating it contains new release: `release-x.y.z`.
```
so that we are consistent :)

PUBLISHING
```diff
@@ -0,0 +1,25 @@
...
+4. Push the change, open PR and wait for approval.
+
+## Github workflow:
+
+1. Open releases dialog in repo main page.
+2. Click on "Draft new release" button.
+3. Create new tag for your release, if there isn't one already. Tag must begin with characte "v".
+4. Set name for your release to match the tag.
+5. Generate release notes or supply them manually.
+6. Upload build artefacts, at very least the wheel build.
```
TomasTomecek, line 23: is this necessary when the release is available on PyPI?

TomasTomecek:
Very nice! Especially the detailed documentation

PUBLISHING
jpodivin: No. But it is better to have backups in another system. The developer should build the package locally before release anyway. Just to be sure.

TomasTomecek: okay, got it

TomasTomecek (approved):

PR #43, opened 2024-07-16
Print friendly errors instead of traceback

xsuchy:
2 commits, see their descriptions individualy.

### DIFF
--- a/logdetective/logdetective.py
+++ b/logdetective/logdetective.py
@@ -66,7 +66,12 @@ def main():
 
     LOG.info("Getting summary")
 
-    log = retrieve_log_content(args.file)
+    try:
+        log = retrieve_log_content(args.file)
+    except ValueError as e:
+        # file does not exists
+        LOG.error(e)
+        sys.exit(4)
     log_summary = extractor(log)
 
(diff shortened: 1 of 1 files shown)


logdetective/logdetective.py
```diff
@@ -49,8 +49,13 @@ def main():
     LOG.setLevel(log_level)
 
     # Primary model initialization
-    model = initialize_model(args.model, filename_suffix=args.filename_suffix,
-                             verbose=args.verbose > 2)
+    try:
+        model = initialize_model(args.model, filename_suffix=args.filename_suffix,
+                                 verbose=args.verbose > 2)
+    except ValueError as e:
```
jpodivin, line 55: Do you think we could match on the `failed to allocate buffer` string somehow? I suppose it ends up in stderr. Otherwise we could go for `Failed to create llama_context`, so we catch only this specific error..

xsuchy: It is hard for me to reproduce it. So I would prefer to keep it as it is.

jpodivin: Ok. That makes sense. We can always come back to this later.

jpodivin (approved):

PR #51, opened 2024-08-06
Run logdetective server in a container

TomasTomecek:


docker-compose.yaml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+services:
+  llama-cpp:
+    build:
+      context: .
+    hostname: llama-cpp-server
+    command: "python3 -m llama_cpp.server --model /models/mistral-7b-instruct-v0.2.Q4_K_M.gguf --host 0.0.0.0 --port 8090"
```
jpodivin, line 7: Do you think we could factor out ports etc. into `.env` file? [0]
It would simplify maintenance of the entire setup, although it's obviously an optional thing.

[0]https://docs.docker.com/compose/compose-file/12-interpolation/

Containerfile
```diff
@@ -0,0 +1,17 @@
+FROM fedora:40
+RUN dnf install -y fastapi-cli python3-fastapi python3-requests python3-drain3 python3-pip python3-pydantic-settings python3-starlette+full \
+    && pip3 install sse-starlette starlette-context huggingface_hub[cli] \
+    && mkdir /src
+
+# we need to bind mount models: this takes a lot of time to download and makes the image huge
+RUN mkdir /models \
+    && huggingface-cli download TheBloke/Mistral-7B-Instruct-v0.2-GGUF mistral-7b-instruct-v0.2.Q4_K_M.gguf --local-dir /models --local-dir-use-symlinks False
```
jpodivin, line 8: If we were distributing this, one could argue for a benefit of packaging both runtime and model together, like llamafile does it. But in this case we should just drop the whole layer and make a mount instead.

docker-compose.yaml
```diff
@@ -0,0 +1,24 @@
...
+    stdin_open: true
+    tty: true
```
jpodivin, line 9:
```
volumes:
  - {MODELS_PTH:-./models}:/models
```

docker-compose.yaml
TomasTomecek: sure, I can do that

docker-compose.yaml
TomasTomecek: interesting, that is some new syntax I'm not aware of

jpodivin: Compose spec allows for simplified templating. This way it's possible to separate structure of the compose, from the variables. So you can override them when you need it. 

For example you can set TTY and stdin to be false by default, but override it CLI to do some debugging. 

There is bunch of other ways you can mess with composes, but I found this one to be the most useful.

Containerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# Fedora's llama-cpp-python is segfaulting on the mistral model above :/
+RUN dnf install -y gcc gcc-c++ python3-scikit-build \
```
jkonecny12, line 11: The same here.

Containerfile
```diff
@@ -0,0 +1,17 @@
+FROM fedora:40
+RUN dnf install -y fastapi-cli python3-fastapi python3-requests python3-drain3 python3-pip python3-pydantic-settings python3-starlette+full \
+    && pip3 install sse-starlette starlette-context huggingface_hub[cli] \
+    && mkdir /src
```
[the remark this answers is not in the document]
TomasTomecek: thanks, good point!

TomasTomecek:
@jpodivin @jkonecny12 thanks for your reviews!

Please check again :)

docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+  llama-cpp:
+    build:
+      context: .
+    hostname: "${LLAMA_CPP_HOST}"
+    command: "python3 -m llama_cpp.server --model ${MODEL_FILEPATH} --host 0.0.0.0 --port ${LLAMA_CPP_SERVER_PORT}"
+    stdin_open: true
+    tty: true
+    env_file: .env
+    ports:
+      - "${LLAMA_CPP_SERVER_PORT-8090}:${LLAMA_CPP_SERVER_PORT-8090}"
```
jpodivin, line 12: There should be `:` here. And for other defaults as well.

docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
+version: "3"
+services:
+  llama-cpp:
+    build:
+      context: .
```
jpodivin, line 5: Podman build fails with
```
  File "/usr/bin/podman-compose", line 33, in <module>
    sys.exit(load_entry_point('podman-compose==1.1.0', 'console_scripts', 'podman-compose')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/podman_compose.py", line 3308, in main
    asyncio.run(async_main())
  File "/usr/lib64/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/podman_compose.py", line 3304, in async_main
    await podman_compose.run()
  File "/usr/lib/python3.12/site-packages/podman_compose.py", line 1634, in run
    retcode = await cmd(self, args)
              ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/podman_compose.py", line 2314, in compose_up
    if await compose.commands["build"](compose, build_args) != 0:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/podman_compose.py", line 2271, in compose_build
    s = await t
        ^^^^^^^
  File "/usr/lib64/python3.12/asyncio/tasks.py", line 631, in _wait_for_one
    return f.result()  # May raise f.exception().
           ^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/podman_compose.py", line 2224, in build_one
    raise OSError("Dockerfile not found in " + ctx)
OSError: Dockerfile not found in /home/jpodivin/repos/logdetective
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x7f695ed8f880>
  File "/usr/lib64/python3.12/asyncio/base_subprocess.py", line 126, in __del__
  File "/usr/lib64/python3.12/asyncio/base_subprocess.py", line 104, in close
  File "/usr/lib64/python3.12/asyncio/unix_events.py", line 568, in close
  File "/usr/lib64/python3.12/asyncio/unix_events.py", line 592, in _close
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 795, in call_soon
  File "/usr/lib64/python3.12/asyncio/base_events.py", line 541, in _check_closed
RuntimeError: Event loop is closed
Error: executing /usr/bin/podman-compose up: exit status 1
```

You should explicitly set the container file here [1]

[1]https://docs.docker.com/compose/compose-file/build/#dockerfile

docker-compose.yaml
```diff
@@ -0,0 +1,27 @@
...
+    stdin_open: true
+    tty: true
+    env_file: .env
+    ports:
+      - "${LLAMA_CPP_SERVER_PORT-8090}:${LLAMA_CPP_SERVER_PORT-8090}"
+    volumes:
+      - ${MODELS_PATH-./models}:/models:Z
+  server:
+    build:
+      context: .
```
jpodivin, line 17: ditto

.env
```diff
@@ -0,0 +1,7 @@
+LLAMA_CPP_SERVER_PORT=8090
+LLAMA_CPP_HOST=llama-cpp-server
+LLAMA_CPP_SERVER=http://$LLAMA_CPP_HOST
+LOGDETECTIVE_SERVER_PORT=8080
```
jpodivin, line 4: Ports here and in the python code should match. Although it's not strictly necessary.

jpodivin (changes requested):

docker-compose.yaml
TomasTomecek: I need to try podman-compose

docker-compose.yaml
TomasTomecek: I'd appreciate if you explained why the ":" should be there. Both are valid syntax.

> ${VAR:-default} -> value of VAR if set and non-empty, otherwise default
> ${VAR-default} -> value of VAR if set, otherwise default

I don't have strong preference either way because no one will touch this anyway.

jpodivin: Well, an empty string will probably break stuff if supplied as host parameter, or port number. It's true that that nobody will override these.

jkonecny12:
I would recommend to move all the container stuff (`Containerfile`, `.env` and `docker-compose.yml`) to the separate directory. Especially the `.env` in the repo root might be harder to find out why it is there.

TomasTomecek:
> I would recommend to move all the container stuff (`Containerfile`, `.env` and `docker-compose.yml`) to the separate directory.

Can you please elaborate? You are essentially asking me to rewrite this PR without providing substantial argument why.

> Especially the `.env` in the repo root might be harder to find out why it is there.

I found out pretty quickly why it's there:

```
$ git grep '\.env'
docker-compose.yaml:    env_file: .env
docker-compose.yaml:    env_file: .env
```

jkonecny12:
It was just my comment (and you can say wish) to keep main directory clean and use subdirectories to separate usage.
Ideally I would like to have a project in the repo root to contain only Makefile (if required) and dot files or configuration files. Everything what could be moved should be moved.

However, I don't insist on that, it's fine in a way as it is. For that reason it was only a comment not review :). Feel free to merge it as it is.

PR #52, opened 2024-08-06
stream the output

xsuchy:
so it is more interactive


xsuchy:
Currently, the logdetective waits a dozen minutes and then prints the result én bloc. I wanted to turn on the streaming, but with streaming on, the models continue answering and never finishes. 
I tried everything possible, but it still does not work.
Does anyone have idea how to stop at the right time?

xsuchy:
I made the stream default, but added --no-stream option to workaround broken models like llama3. But instead of heuristics when not to keep streaming on, I choose rather to document the broken model in README.

Ready for re-review.

TomasTomecek (approved):
Tested locally and works really well!

Although there is a big gap (~30 seconds) after the 'Explanation:' text is printed and before the output starts appearing. We may investigate what's happening (is the model being loaded to memory and initialized?) and print some "progress bar" there.

LGTM, very nice!

README.md
```diff
@@ -55,6 +55,10 @@ Example you want to use a different model:
     logdetective https://example.com/logs.txt --model https://huggingface.co/QuantFactory/Meta-Llama-3-8B-Instruct-GGUF/resolve/main/Meta-Llama-3-8B-Instruct.Q5_K_S.gguf?download=true
     logdetective https://example.com/logs.txt --model QuantFactory/Meta-Llama-3-8B-Instruct-GGUF
 
+Note that streaming with some models (notably Meta-Llama-3 is broken) is brokend and can be workarounded by `no-stream` option:
```
jpodivin, line 58: s/brokend/broken/

jpodivin, line 58: It's not just streaming that is broken. That model is broken in general. If you run it without streaming it just keeps on printing into buffer, and output will never show up.

xsuchy: Weird, it shows for me with --no-stream. Althought the delays is about dozen minutes.

xsuchy: Fixed.

jpodivin (approved):

PR #54, opened 2024-08-07
Excluding broken llama-cpp-python version and setting new release

jpodivin:


jkonecny12 (approved):
LGTM!

jkonecny12 (changes requested):
Tested locally and still getting the issue:
```

Note: This error originates from the build backend, and is likely not a problem with poetry but with llama-cpp-python (0.2.85) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "llama-cpp-python (==0.2.85)"'.
```

jkonecny12 (approved):
It works at the end. Issue on my end was wrong local dependencies. Resolved.

PR #56, opened 2024-08-19
add CUDA container for logdetective

TomasTomecek:
this enables running logdetective in a container while accelerated by CUDA

details: https://blog.tomecek.net/post/running-logdetective-service-on-cuda-ec2/

### DIFF
--- a/Containerfile.cuda
+++ b/Containerfile.cuda
@@ -1,0 +1,19 @@
+FROM fedora:39
+# make sure nvidia driver match on host and in the container
+RUN dnf install -y python3-requests python3-pip gcc gcc-c++ python3-scikit-build git-core \
+    && echo "[cuda-fedora39-x86_64]" >> /etc/yum.repos.d/cuda.repo \
+    && echo "name=cuda-fedora39-x86_64" >> /etc/yum.repos.d/cuda.repo \
+    && echo "baseurl=https://developer.download.nvidia.com/compute/cuda/repos/fedora39/x86_64" >> /etc/yum.repos.d/cuda.repo \
+    && echo "enabled=1" >> /etc/yum.repos.d/cuda.repo \
+    && echo "gpgcheck=1" >> /etc/yum.repos.d/cuda.repo \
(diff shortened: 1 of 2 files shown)


jkonecny12:
I wonder, could we somehow use RHEL AI? Isn't there a Fedora AI variant?

Containerfile.cuda
```diff
@@ -0,0 +1,19 @@
+FROM fedora:39
+# make sure nvidia driver match on host and in the container
+RUN dnf install -y python3-requests python3-pip gcc gcc-c++ python3-scikit-build git-core \
+    && echo "[cuda-fedora39-x86_64]" >> /etc/yum.repos.d/cuda.repo \
+    && echo "name=cuda-fedora39-x86_64" >> /etc/yum.repos.d/cuda.repo \
+    && echo "baseurl=https://developer.download.nvidia.com/compute/cuda/repos/fedora39/x86_64" >> /etc/yum.repos.d/cuda.repo \
+    && echo "enabled=1" >> /etc/yum.repos.d/cuda.repo \
+    && echo "gpgcheck=1" >> /etc/yum.repos.d/cuda.repo \
+    && echo "gpgkey=https://developer.download.nvidia.com/compute/cuda/repos/fedora39/x86_64/D42D0685.pub" >> /etc/yum.repos.d/cuda.repo \
```
jkonecny12, line 9: Wouldn't it be better to just have .repo file next to the Containerfile and COPY it into the container. It's seems simpler.

xsuchy (approved):
+1

Containerfile.cuda
TomasTomecek: yes it would, but your [comment](https://github.com/fedora-copr/logdetective/pull/51#issuecomment-2277488054) would become even more relevant :) I agree I don't want to flood the root dir of this repo with "random" files

TomasTomecek:
> I wonder, could we somehow use RHEL AI? Isn't there a Fedora AI variant?

since we discussed this on our sync meeting, I'll reply here as well

there is no official release right now of those: once something's out, I totally want to explore :)

Containerfile.cuda
TomasTomecek: I really like that now we can see the content of the repo file and i.e. once F40 variant is available, we can `s/39/40/g` in one file

PR #58, opened 2024-08-29
alway add newline at the of stream

xsuchy:
Sometimes the answer from model does not end with newline. When we stream we do not add newline so the command line can be broken. This does not affect no-stream flow.


jpodivin (approved):

TomasTomecek (approved):
nice catch!

PR #59, opened 2024-09-02
Adding general response certainty based on median log probability

jpodivin:
Please manually test before approving.


logdetective/utils.py
```diff
@@ -1,10 +1,12 @@
 import logging
 import os
+from typing import Iterator
 from urllib.parse import urlparse
-
+import numpy as np
```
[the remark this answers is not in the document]
TomasTomecek: This should install numpy in the GHA environment:
```
- name: VCS Diff Lint
  uses: fedora-copr/vcs-diff-lint-action@v1
  id: VCS_Diff_Lint
  install_rpm_packages: ["python3-numpy"]
```

TomasTomecek (approved):
Tried this locally on a log from our deployment with some python tracebacks and other errors.
```
The root cause of the problem might be related to the Copr configuration, such as an incorrect or missing `/etc/copr/build_chroot.d/` directory or incorrect Copr client setup.

To resolve these issues, you may consider the following steps:

1. Ensure that the required directory, '/persistent/results', exists and is accessible by the application.
2. Check Copr configuration and ensure that it is properly set up. This might include ensuring that the `/etc/copr/build_chroot.d/` directory exists and contains the necessary configuration files.
3. Make sure that the Copr client is installed and configured correctly.
4. If necessary, try to recreate the build environment to see if the issues persist.

Once the above steps have been taken, re-run the build process to see if the errors have been resolved. If the issues still persist, further investigation may be necessary to determine the root cause.
Response certainty: 99.9976277914696%
```
Logdetective was pretty certain of its response

logdetective/utils.py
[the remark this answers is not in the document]
jpodivin: It's there.

logdetective/logdetective.py
```diff
@@ -92,15 +93,21 @@ def main():
...
         # Stream the output
         for chunk in response:
+            if isinstance(chunk["choices"][0]["logprobs"], dict):
+                probs.extend(chunk["choices"][0]["logprobs"]["top_logprobs"])
             delta = chunk['choices'][0]['text']
             print(delta, end='', flush=True)
-        print()
+    certainty = compute_certainty(probs)
+
+    print(f"\nResponse certainty: {certainty:.2f}%\n")
```
xsuchy, line 110: I think we can go low as 1f or even 0f.

pyproject.toml
```diff
@@ -31,6 +31,7 @@ requests = "^2.31.0"
 llama-cpp-python = "^0.2.56,!=0.2.86"
 drain3 = "^0.9.11"
 huggingface-hub = ">0.23.2"
+numpy = "^2.1.0"
```
xsuchy, line 34: Fedora 40 has numpy 1.26.4. Both `median` and `exp` are present there. So `>= 1.26.4' is ok. The API interface changed last time in 2008 (about 1.2 release). `exp` was hard to find. But I believe it is even older. Therefore I recommend removing any version requirement and require simply `numpy`.

logdetective/utils.py
```diff
@@ -57,20 +58,32 @@ def initialize_model(model_pth: str, filename_suffix: str = ".gguf", verbose: bo
...
 
 
-def process_log(log: str, model: Llama, stream: bool) -> str:
+def compute_certainty(probs: List[Dict[str, float] | None]) -> float:
+    """Compute certainty of repsponse based on average logit probability
     """
-    Processes a given log using the provided language model and returns its summary.
+
+    top_logprobs = [
+        np.exp(x) * 100 for e in probs if isinstance(e, dict) for x in e.values()]
```
xsuchy, line 79: I will welcome comment on why this formula.

xsuchy:
Otherwise LTGM and very nice :clap:

logdetective/logdetective.py
jpodivin: I'm sure we can, although I kind of like the two decimal cut off.

logdetective/utils.py
jpodivin: Log probability is `log(p)`, isn't really readable for most people, especially in compound. IT needs to be converted first. 

In this case it's just a  matter of applying inverse operation `exp`. Of course that leaves you with a value in range `<0, 1>` so it needs to be multiplied by `100`.

xsuchy: Can you put these two lines as comment there. It will be helpful for any of us fixing this code 5 years in the future.

logdetective/logdetective.py
xsuchy: Then leave it there. We will ask users later how they like it.

Issue #61, opened 2024-09-03
Fix code scanning alert - Full server-side request forgery


jpodivin:
Tracking issue for:
- [ ] https://github.com/fedora-copr/logdetective/security/code-scanning/37

PR #62, opened 2024-09-03
Simple validation of URL from request

jpodivin:
I really didn't want to use regex for this. Or a whitelist of allowed log sources, since we don't know what will people use. 
This way we can avoid at leas some suspicious values, those without right scheme and unnecessary parts. 

Should resolve: #61


TomasTomecek (approved):
Looks good :+1:

Issue #60, opened 2024-09-03
Model re-download after we changed the default to our own


TomasTomecek:
bd5c0cddc2b82f5b0c841d687bc6ea7765c8d2c6

```
$ logdetective -v some.log
ggml-model-Q4_K_S.gguf: 100%|█████████████████████████████| 4.14G/4.14G [05:42<00:00, 10.7MB/s]
```

I was frustrated I had to download the model again.

Now that I look how HF stores models locally, it seems it has to be downloaded anyway:
```
$ ll ~/.cache/huggingface/hub/models--fedora-copr--Mistral-7B-Instruct-v0.2-GGUF/snapshots/f8d679d67ad3e551f131ceff60a673d3a6ff4286/ggml-model-Q4_K_S.gguf
lrwxrwxrwx. 1 tt tt 76 sep  3 09:58 /home/tt/.cache/huggingface/hub/models--fedora-copr--Mistral-7B-Instruct-v0.2-GGUF/snapshots/f8d679d67ad3e551f131ceff60a673d3a6ff4286/ggml-model-Q4_K_S.gguf -> ../../blobs/be98b5d472de622d4d413550e00d04b06c35e7bec087ddb1f10837687c735606

$ ll ~/.cache/huggingface/hub/models--TheBloke--Mistral-7B-Instruct-v0.2-GGUF/snapshots/3a6fbf4a41a1d52e415a4958cde6856d34b2db93/mistral-7b-instruct-v0.2.Q4_K_S.gguf
lrwxrwxrwx. 1 tt tt 76 aug  8 10:53 /home/tt/.cache/huggingface/hub/models--TheBloke--Mistral-7B-Instruct-v0.2-GGUF/snapshots/3a6fbf4a41a1d52e415a4958cde6856d34b2db93/mistral-7b-instruct-v0.2.Q4_K_S.gguf -> ../../blobs/1213e19b3e103932fdfdc82e3b6dee765f57ad5756e0f673e7d36514a5b60d0a
```

It would be amazing if the filename at least said it's mistral.

jpodivin:
I agree that it's unfortunate the model is duplicated this way. But it's actually necessary. We can compare models layer by layer, to ascertain they are same or not, but that take time, hash is fast, but it will be affected by even minute changes in metadata, which can happen between llama.cpp releases. 

[...]

PR #63, opened 2024-09-04
Log model location when it's being pulled

jpodivin:
Regarding: #60

It prints out the location, on the info level, that's the level I see as most fitting.


TomasTomecek (approved):
Thank you! Perfect

PR #65, opened 2024-09-05
Setting up server API

jpodivin:
The server now has properly defined API, with data model of response. Error handling was implemented for various failure modes, including possibility of invalid log probabilities.

### DIFF
--- a/logdetective/server.py
+++ b/logdetective/server.py
@@ -1,20 +1,46 @@
+import json
 import logging
 import os
-import json
+from typing import List
 
+from llama_cpp import CreateCompletionResponse
 from fastapi import FastAPI, HTTPException
 from pydantic import BaseModel
 
 import requests
 
-from logdetective.constants import PROMPT_TEMPLATE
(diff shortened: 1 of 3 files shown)


logdetective/server.py
```diff
@@ -1,21 +1,29 @@
...
 class BuildLog(BaseModel):
     """Model of data submitted to API.
     """
     url: str
 
+
+class Response(BaseModel):
+    """Model of data returned by Log Detective API
+    """
+    llm_response: CreateCompletionResponse
```
TomasTomecek, line 23: Can we see an example content of this?

jpodivin: Sure, it's all part of llama-cpp-python API https://llama-cpp-python.readthedocs.io/en/latest/api-reference/#llama_cpp.llama_types.CreateCompletionResponse

TomasTomecek: nice, that's pretty detailed

please link it in the code so it's easy to find :)

TomasTomecek (approved):
very nice, we're getting there with the service \o/

logdetective/server.py
jpodivin: Done

PR #73, opened 2024-10-04
Adding publishing workflow

jpodivin:


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,112 @@
...
+      # https://github.com/docker/build-push-action
+      - name: Build CUDA image
+        id: build-image-cuda
+        uses: docker/build-push-action@0565240e2d4ab88bba5387d719585280857ece09 # v5.0.0
+        with:
+          context: .
+          file: Containerfile.cuda
+          push: ${{ github.event_name != 'pull_request' && contains(github.ref, 'refs/tags/') }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
TomasTomecek, line 95: i think we may need to append `-cuda` somewhere here to the image name

TomasTomecek (approved):
Thanks!

This looks like it could work, let's give it a go.

PR #76, opened 2024-10-07
api: require Authentication header when configured

TomasTomecek:
so we can run it securely without being bombed by bots

if LOGDETECTIVE_TOKEN is not set, the authentication check is disabled
for sake of local development


logdetective/server.py
```diff
@@ -44,13 +44,30 @@ class StagedResponse(Response):
...
 
 LLM_CPP_HOST = os.environ.get("LLAMA_CPP_HOST", "localhost")
 LLM_CPP_SERVER_ADDRESS = f"http://{LLM_CPP_HOST}"
 LLM_CPP_SERVER_PORT = os.environ.get("LLAMA_CPP_SERVER_PORT", 8000)
 LLM_CPP_SERVER_TIMEOUT = os.environ.get("LLAMA_CPP_SERVER_TIMEOUT", 600)
 LOG_SOURCE_REQUEST_TIMEOUT = os.environ.get("LOG_SOURCE_REQUEST_TIMEOUT", 60)
+API_TOKEN = os.environ.get("LOGDETECTIVE_TOKEN", None)
+
+def requires_token_when_set(authentication: Annotated[str | None, Header()] = None  ):
+    if not API_TOKEN:
```
jpodivin, line 56: Could we log this process?

TomasTomecek: oh yeah, very good point!

logdetective/server.py
```diff
@@ -44,13 +44,42 @@ class StagedResponse(Response):
...
+        except (ValueError, IndexError):
+            LOG.warning(
+                "Authentication header has invalid structure (%s), it should be 'Bearer TOKEN'",
+                authentication)
+            # eat the exception and raise 401 below
+            token = None
+        if token == API_TOKEN:
+            return
+    LOG.info("LOGDETECTIVE_TOKEN env var is set (%s), clien token = %s",
+             API_TOKEN, token)
```
jpodivin, line 79: Just maybe not print the token value.

TomasTomecek: Why not? these are our logs; if someone can see that log, they can certainly access podman (and see the env var set there) or have shell access already.

Printing both values will make debugging much easier in case of issues.

jpodivin: Well if that's ok with you I'm not going to hold this PR over it.

jpodivin (approved):

PR #80, opened 2024-10-15
Parametrizing number of offloaded layers

jpodivin:
Handful of changes necessary to get the GPU offloading to work by default. The commented out parts are necessary as well, but this way it will still work on CPU only systems.


TomasTomecek (approved):
LGTM

PR #81, opened 2024-10-17
Improve server part of the README.md

jkonecny12:
For easy setup of the server users have to have model locally downloaded, however, we are missing easy guide to know which model should be used or how to download it.

Let's make things a bit more accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,11 @@ $ ll $MODELS_PATH
 
 If the variable is not set, `./models` is mounted inside by default.
 
+Model can be downloaded by:
+```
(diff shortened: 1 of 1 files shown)


jpodivin (commented):
Thanks, but could you also add link to our space on Hugging Face? I plan to put our models in there in the long run, so it would be good to have it here.

jkonecny12:
Is this enough?

TomasTomecek (approved):
Nice, this is very clever, no need of dedicated tooling, just plain & simple `curl`

jpodivin (approved):

PR #83, opened 2024-10-25
Streaming end point for server API

jpodivin:
Implements new endpoint for streaming model outputs.

Responses follow the same structure as the existing API, with a caveat, that each token has it's own structure, with probability etc. returned separately, as part of the stream.

### DIFF
--- a/logdetective/server.py
+++ b/logdetective/server.py
@@ -5,8 +5,8 @@ from typing import List, Annotated
 
 from llama_cpp import CreateCompletionResponse
 from fastapi import FastAPI, HTTPException, Depends, Header
+from fastapi.responses import StreamingResponse
 from pydantic import BaseModel
-
 import requests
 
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Looks good to me :+1:

logdetective/server.py
```diff
@@ -129,32 +129,36 @@ def submit_text(text: str, max_tokens: int = 0, log_probs: int = 1):
...
+                    f"[{response.status_code}] {response.text}")
+        try:
+            response = json.loads(response.text)
+        except UnicodeDecodeError as ex:
+            LOG.error("Error encountered while parsing llama server response: %s", ex)
+            raise HTTPException(
+                status_code=400,
+                detail=f"Couldn't parse the response.\nError: {ex}\nData: {response.text}") from ex
+    else:
+        return response
```
TomasTomecek, line 161: I'm sure we'll need to tinker with the error handling for the streaming response in future once we'll start using this.

TomasTomecek (approved):
Very nice!

PR #92, opened 2024-12-02
Adding flake8 hook and checking all files

jpodivin:


TomasTomecek (approved):
LGTM, very good suggestion!

How about black/ruff - I love using those so we don't need to discuss code formatting :)

jpodivin:
We could do ruff as well. I'm wondering how is the hook integration though. If people wouldn't need to download another binary separately.

praiskup:
>  I'm wondering how is the hook integration though

Not really related to this particular PR, but we are using ruff using vcs-diff-lint in Copr:
https://github.com/fedora-copr/copr/blob/main/.github/workflows/python-diff-lint.yml#L34-L36

PR #98, opened 2025-01-06
utilize both numpy 1.26 and 2.0 + use Packit to build RPMs and run tests for PRs

TomasTomecek:
stable Fedora is on 1.26 while rawhide already has 2.2
we need to support both

the function compute_certainty is not used in the CLI codebase but in
server

we need this to properly build detective in Fedora as an RPM


jkonecny12:
Hi, shouldn't this be split into multiple commits? It seems to me that some changes should be separated?

xsuchy:
I can live with one commit, but these lint issues would be nice to resolve.
otherwise LGTM

TomasTomecek:
> Hi, shouldn't this be split into multiple commits? It seems to me that some changes should be separated?

I'm sorry but I don't follow what exactly should I be splitting. In this one single commit I verified the current codebase works with numpy 1 and 2 thanks to the new test case I created. Please suggest which changes should land in a separate commit and what should be the title.

TomasTomecek:
I will update this PR by the end of the week, thanks for reviews

jkonecny12 (approved):
Looks good to me now. Thanks!

.packit.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - poetry build -f sdist
+    - bash -c "ls -1t ./dist/*.tar.gz | head -n 1"
+  get-current-version:
+    - poetry version -s
+
+jobs:
+  - job: propose_downstream
+    trigger: release
+    dist_git_branches:
+      - fedora-all
```
xsuchy, line 22: We do not want to push to all branches. The deps will be very likely broken in old branches.

.packit.yaml
```diff
@@ -0,0 +1,40 @@
...
+    # rawhide specfile will be the source of truth: we don't need to have the specfile upstream
+    - curl -Lvso python-logdetective.spec https://src.fedoraproject.org/rpms/python-logdetective/raw/main/f/python-logdetective.spec
+  create-archive:
+    - poetry build -f sdist
+    - bash -c "ls -1t ./dist/*.tar.gz | head -n 1"
+  get-current-version:
+    - poetry version -s
+
+jobs:
+  - job: propose_downstream
```
xsuchy, line 19: Do we want it? I have to admit that I am more happier with pull-from-upstream.

TomasTomecek: good point, I'll merge both packit.yaml configs together

.packit.yaml
TomasTomecek: dropped propose_downstream in favor of pull_from_upstream

jpodivin (approved):

PR #102, opened 2025-01-09
run our server with TLS

TomasTomecek:
we're using the same website certificates

at some point, we may need to add nginx as well based on recommendation fron gunicorn docs: https://docs.gunicorn.org/en/latest/deploy.html

production deployment now has a config and smaller timeout because we run inference on a GPU

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+/server/fullchain.pem
+/server/privkey.pem
+/server/cert.pem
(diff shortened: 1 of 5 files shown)


TomasTomecek:
I can't fix those because it's a config for gunicorn and they need to be named like that

jpodivin (approved):

jpodivin:
The diff-lint seems to have an issue with a constant name. If it can't be renamed I suggest we add the file to ignorelist.

Issue #108, opened 2025-01-14
Paralel processing of /explain/


xsuchy:
Today, we are able to process only one request to /explain in a time. This does not scale.

We should investigate how to parallelize this.

In https://github.com/fedora-copr/logdetective/pull/106/files I discovered that:

I found that llama_cpp.server should have --parallel 2.See https://www.reddit.com/r/LocalLLaMA/comments/1be845y/multiple_concurrent_generations_with_llamacpp/
Is this what we need for parallel access?

But later @tt discovered that:

> unfortunately, we are using llama-cpp-python's server and not the one in llama.cpp :// we are using this one and it doesn't know the parallel flag - I saw only threading being configurable https://github.com/abetlen/llama-cpp-python?tab=readme-ov-file#openai-compatible-web-server

We should either replace our server or find another way how to parallelize this.

jpodivin:
Replacement with the server from llama.cpp would be the most straightforward.

TomasTomecek:
We are still running it sequentially but I'll increase the parallelism today now that we have llama-server up.

Just tried 2 parallel queries and got this:
```
[2025-01-23 09:31:33 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:16)
[2025-01-23 09:31:34 +0000] [1] [ERROR] Worker (pid:16) was sent code 134!
[2025-01-23 09:31:34 +0000] [22] [INFO] Booting worker with pid: 22
[2025-01-23 09:31:35 +0000] [22] [INFO] Started server process [22]
[2025-01-23 09:31:35 +0000] [22] [INFO] Waiting for application startup.
[2025-01-23 09:31:35 +0000] [22] [INFO] Application startup complete.
[2025-01-23 09:31:45 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:12)
[2025-01-23 09:31:45 +0000] [1] [ERROR] Worker (pid:12) was sent code 134!
[2025-01-23 09:31:45 +0000] [26] [INFO] Booting worker with pid: 26
[2025-01-23 09:31:46 +0000] [26] [INFO] Started server process [26]
[2025-01-23 09:31:46 +0000] [26] [INFO] Waiting for application startup.
[2025-01-23 09:31:46 +0000] [26] [INFO] Application startup complete.
```

so I may also need to increase gunicorn timeouts

TomasTomecek:
This is now implemented and deployed; I tested it on 2 logs processed in parallel and it worked flawlessly.

[...]

PR #111, opened 2025-01-17
Moving to llama.cpp server

jpodivin:
Should address: https://github.com/fedora-copr/logdetective/issues/108


.env
```diff
@@ -3,8 +3,14 @@
...
 PYTHONPATH=/src:/usr/local/lib64/python3.12/site-packages:/usr/lib64/python312.zip:/usr/lib64/python3.12/:/usr/lib64/python3.12/lib-dynload:/usr/local/lib/python3.12/site-packages:/usr/lib64/python3.12/site-packages:/usr/lib/python3.12/site-packages
-LLM_NGPUS=-1
-LLAMA_CPP_CONFIG=llama_cpp_server_config.json
+LLAMA_ARG_MODEL="/models/mistral-7b-instruct-v0.2.Q4_K_S.gguf"
+LLAMA_ARG_ALIAS="default-model"
+LLAMA_ARG_N_GPU_LAYERS=-1
+LLAMA_ARG_THREADS=12
+LLAMA_ARG_BATCH=512
+# Modify following var when switching model
+LLAMA_ARG_CHAT_TEMPLATE="mistral-v3"
```
TomasTomecek, line 14: are these compatible? v3 vs. our v0.2 model?

.env
```diff
@@ -3,8 +3,14 @@
...
+LLAMA_ARG_CTX_SIZE=32768
+LLAMA_ARG_N_PARALLEL=4
```
TomasTomecek, line 16: I'll tinker with this and may start with just 2 and see the performance

Containerfile.cuda
```diff
@@ -8,12 +8,22 @@ RUN dnf install -y python3-requests python3-pip gcc gcc-c++ python3-scikit-build
...
+
+# Clone, checkout, build and move llama.cpp server to path
+RUN git clone https://github.com/ggerganov/llama.cpp.git && \
+    cd llama.cpp && \
+    git checkout $LLAMACPP_VER && \
+    cmake -B build && \
+    cmake --build build --config Release -j 4 -t llama-server && \
+    mv ./build/bin/llama-server /bin/llama-server
+
+# COPY ./logdetective/ /src/logdetective/logdetective
```
TomasTomecek, line 29: good point, we don't need this for the llama server :)

TomasTomecek (approved):
good work, will deploy it on Mon and will also set up the GHA for image building

Containerfile.cuda
jpodivin: Actually, I should just delete it.

.env
jpodivin: Yes, the chat templates are actually identical in version 2 and 3.
The tokenizer has been slightly modified, but not in a way that would affect our use case.
https://github.com/mistralai/cookbook/blob/main/concept-deep-dive/tokenization/chat_templates.md

Containerfile.cuda
```diff
@@ -8,12 +8,20 @@ RUN dnf install -y python3-requests python3-pip gcc gcc-c++ python3-scikit-build
...
+#     pydantic-settings fastapi[standard] \
+#     && mkdir /src
+
+# Clone, checkout, build and move llama.cpp server to path
+RUN git clone https://github.com/ggerganov/llama.cpp.git && \
+    cd llama.cpp && \
+    git checkout $LLAMACPP_VER && \
+    cmake -B build && \
+    cmake --build build --config Release -j 4 -t llama-server && \
+    mv ./build/bin/llama-server /bin/llama-server
```
xsuchy, line 27: Should this be /usr/bin/ ?

jpodivin: I think it's just link to /usr/bin anyway, and both are in path so it doesn't make any difference.

TomasTomecek:
Merging so I can give it a go.

Containerfile.cuda
xsuchy: Yes. But since the merge, we should use the correct path. And that is /usr/bin
https://discussion.fedoraproject.org/t/f40-change-proposal-unify-usr-bin-and-usr-sbin-system-wide/99853

jpodivin: Right. I will do that in a followup.

PR #118, opened 2025-01-23
push container images for every commit

TomasTomecek:
I am building them now locally myself, this would help us a lot

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -35,10 +35,10 @@ jobs:
       # Install and test the cosign tool except on PR
       # https://github.com/sigstore/cosign-installer
       - name: Install cosign
-        if: ${{ github.event_name != 'pull_request' && contains(github.ref, 'refs/tags/') }}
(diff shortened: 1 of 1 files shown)


jpodivin:
I don't think this is something we should do. Releasing a new image should be a deliberate action, with release notes and tags involved.

jpodivin (commented):
I don't think this is something we should do. Releasing a new image should be a deliberate action, with release notes and tags involved.

sgallagher:
> I don't think this is something we should do. Releasing a new image should be a deliberate action, with release notes and tags involved.

What if we push a container image for each commit but only push the `:latest` tag on proper releases? That would allow us to pull development containers between releases.

jpodivin:
> > I don't think this is something we should do. Releasing a new image should be a deliberate action, with release notes and tags involved.
> 
> What if we push a container image for each commit but only push the `:latest` tag on proper releases? That would allow us to pull development containers between releases.

Yes, that would also work.

PR #122, opened 2025-02-27
Server config

jpodivin:
Server behavior can now be controlled by configuration file.

The server code was refactored, to simplify work with configuration file and open options for further parametrization. 

Since I'm moving the server.py, and changing dependencies, this will require changes to the spec file as well.


server/config.yml
```diff
@@ -0,0 +1,7 @@
+inference:
```
TomasTomecek, line 1: love this, much better than flooding the CLI args

TomasTomecek (commented):
I'm sorry I didn't get to this

I'll make it a prio next week and even deploy it so we can test it live

pyproject.toml
```diff
@@ -35,16 +35,18 @@ huggingface-hub = ">0.23.2"
 # we need to support both versions
 numpy = ">=1.26.0"
 
+[tool.poetry.group.server.dependencies]
+fastapi = "^0.115.8"
+pydantic = "^2.10.6"
+pyyaml = "^6.0.2"
```
xsuchy, line 41: Sigh. I really hate this top-version by poetry. Does it means that pyyaml 6.0.1 will not work? 5.4 will not work?

logdetective/server/models.py
```diff
@@ -0,0 +1,82 @@
...
+class Response(BaseModel):
+    """Model of data returned by Log Detective API
+
+    explanation: CreateCompletionResponse
+        https://llama-cpp-python.readthedocs.io/en/latest/api-reference/#llama_cpp.llama_types.CreateCompletionResponse
+    response_certainty: float
+    """
+
+    explanation: Dict
+    response_certainty: float
```
xsuchy, line 20: Do we really want this as class attributes (as opposed to instance attributes). Especially in this class I think this is error. As any change of this attribute change the value will affect all instances of the class.
In all other classes, it can be arguable, but it can be problematic one day. I would recommend using instance attributes as they are safer, and class attributes, in this case, do not bring any benefit.

pyproject.toml
jpodivin: Caret constraints are somewhat more complicated than explicit `><=` but essentially it's just a short hand for the same. In case of pyyaml, `>=6.0.2,<7.0.0`. 

https://python-poetry.org/docs/dependency-specification/#caret-requirements

logdetective/server/models.py
jpodivin: This class is based on pydantic `BaseModel`[1], essentially a struct with field validation, or a "super-dataclass"[2]. So these are not used as class attributes, but instance attributes, set either during initialization or, which isn't a regular usage, on a created instance.

The behavior is then:

```
>>> from pydantic import BaseModel
... 
... 
... class User(BaseModel):
...     id: int
...     name: str = 'Jane Doe'
...     
>>> a = User(id=10)
>>> b = User(id=20)
>>> (a.id, b.id)
(10, 20)
>>> a.id = 35
>>> (a.id, b.id)
(35, 20)
>>> b.id = 0
>>> (a.id, b.id)
(35, 0)
```

Of course, we could set class attribute to a value. But it wouldn't affect any of the instances:

```
>>> User.name = "James Doe"
>>> a.name
'Jane Doe'
>>> User.name
'James Doe'
```
Because these are not really class attributes, although they may look like it.

[1] https://docs.pydantic.dev/latest/concepts/models/
[2] https://docs.python.org/3/library/dataclasses.html#module-dataclasses

.env
```diff
@@ -15,3 +17,5 @@ LLAMA_ARG_BATCH=512
 LLAMA_ARG_CHAT_TEMPLATE="mistral-v3"
 LLAMA_ARG_CTX_SIZE=32768
 LLAMA_ARG_N_PARALLEL=4
+# Path to Logdetective server configuration file
+LOGDETECTIVE_SERVER_CONF="/config.yml"
```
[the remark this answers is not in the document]
jpodivin: These parameters are only used for container service, we already abuse the filesystem pretty heavily there, with `/models` and `/src` but I can move it somewhere else. It's just a question of location. Do you have something in mind?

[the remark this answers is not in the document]
xsuchy: Ack. I will not block on that. But one small kitty will die on every abuse of FS.

pyproject.toml
xsuchy: I know the caret operator. I was arguing if there can be:

* ^6.0.1
* or >5.4
* or >1.0

I think that we (as developers) should spend time investigating what versions we really need. And not use the latest version that is available.

server/config.yml
jpodivin: Thank you!

TomasTomecek (approved):
Just verified it works well locally in containers. Very nice work!

PR #123, opened 2025-03-05
New heuristic for chunk creation

jpodivin:


xsuchy (approved):
+1

logdetective/utils.py
```diff
@@ -15,10 +15,17 @@
...
+    * is the next character is whitespace
+    * is the previous character backslash '\\'
+    * is the previous character colon ':'
+
     """
     conditionals = [
         lambda i, string: string[i + 1].isspace(),
-        lambda i, string: string[i - 1] == "\\"
+        lambda i, string: string[i - 1] == "\\",
+        lambda i, string: string[i - 1] == ":"
```
TomasTomecek, line 28: can you please elaborate what exactly this achieves? I am curious about the use case

jpodivin: The function is described in the the docstring. The intended effect is to pick out cases when log line ends with `:`  but the next line doesn't start with whitespace.

Just like in normal text, `:` indicates that what follows is continuation, elaboration or explanation of the preceding text, which in turn provides some sort of general context, subject or headline. Previously, if split on multiple lines without indentation, these strings were split.

Now we can at least get the first continuation line for sure. The following lines are going to be lost anyway. Unless they fit in any of the other rules. As of now I haven't found a way to ensure it happens in a way that is both reliable and fast.

TomasTomecek: understood, thanks for the detailed explanation

PR #129, opened 2025-03-14
Adding option to provide authentication token for LLM

jpodivin:
If we want to use LLMs from third party providers we need to supply authentication tokens.
OpenAI APIs [1][2] expect this as part of the request header, the same approach was adopted by perplexity [3] and others so it makes sense to support it.

[1]https://platform.openai.com/docs/api-reference/completions
[2]https://platform.openai.com/docs/api-reference/chat/create
[3]https://docs.perplexity.ai/api-reference/chat-completions
[4]https://huggingface.co/docs/api-inference/en/tasks/chat-completion?code=curl#code-snippet-example-for-conversational-llms


TomasTomecek (approved):
very clear, thanks!

PR #132, opened 2025-03-17
sync production configuration to git (registry namespace, gunicorn timeout)

TomasTomecek:
```
increase backend API timeout to 10 minutes

2 minutes is not enough sadly, especially now that we make much more
snippets
```

```
prod compose: point to quay.io images

we need to point to our officially built quay.io images, otherwise
podman treats it as localhost
```

### DIFF
--- a/server/gunicorn-prod.config.py
+++ b/server/gunicorn-prod.config.py
@@ -2,8 +2,9 @@ import os
 bind = f"0.0.0.0:{os.environ['LOGDETECTIVE_SERVER_PORT']}"
 worker_class = "uvicorn.workers.UvicornWorker"
 workers = 2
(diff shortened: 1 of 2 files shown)


server/gunicorn-prod.config.py
```diff
@@ -3,7 +3,7 @@
 worker_class = "uvicorn.workers.UvicornWorker"
 workers = 2
 # timeout set to 120 - 2 minutes should be enough for one LLM execution in production on a GPU
```
jpodivin, line 5: Please update the comment as well.

jpodivin (changes requested):

server/gunicorn-prod.config.py
TomasTomecek: thanks for noticing, updates

jpodivin (approved):

PR #137, opened 2025-03-19
Listen for GitLab job_event webhooks

sgallagher:
Listens for the events and queries the GitLab API for additional
information leading to the logs of failed builds.

Retrieves the logs and analyzes them to determine the correct log to
submit to the Log Detective processing.

Fixes: https://issues.redhat.com/browse/LD-8

### DIFF
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -325,8 +325,13 @@ async def process_gitlab_job_event(job_hook):
         return
     merge_request_id = int(match.group(1))
 
+    LOG.debug("Retrieving log artifacts")
     # Retrieve the build logs from the merge request artifacts and preprocess them
-    preprocessed_log = await retrieve_and_preprocess_koji_logs(job)
+    try:
+        preprocessed_log = await retrieve_and_preprocess_koji_logs(job)
+    except LogsTooLargeError:
+        LOG.error("Could not retrieve logs. Too large.")
+        raise
 
     # Submit log to Log Detective and await the results.
     response = await submit_log_to_llm(preprocessed_log)
@@ -334,6 +339,10 @@ async def process_gitlab_job_event(job_hook):
 
     # Add the Log Detective response as a comment to the merge request
     await comment_on_mr(merge_request_id, response)
+
+
+class LogsTooLargeError(RuntimeError):
+    """The log archive exceeds the configured maximum size"""
 
 
 async def retrieve_and_preprocess_koji_logs(job):
@@ -345,6 +354,12 @@ async def retrieve_and_preprocess_koji_logs(job):
     returns: An open, file-like object containing the log contents to be sent
     for processing by Log Detective. The calling function is responsible for
     closing this object."""
+
+    # Make sure the file isn't too large to process.
+    if not await check_artifacts_file_size(job):
+        raise LogsTooLargeError(
+            f"Oversized logs for job {job.id} in project {job.project_id}"
+        )
 
     # Create a temporary file to store the downloaded log zipfile.
     # This will be automatically deleted when the last reference into it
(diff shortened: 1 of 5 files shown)


sgallagher:
Still to do:

- Decide how we're going to maintain the opt-in list of packages and implement that
- Rework the tempfile implementation. It was originally done with a tempdir when I was doing a full zipfile extraction, but since we're now using the `zipfile` module, it can be simplified.
- Determine if we want to be maintaining the log content entirely in-memory or if we can pass a file-like object into Log Detective as a stream from the `zipfile` interface. If so, revise the tempfile to stay available until full processing is complete.

TomasTomecek:
We need to improve that linter because it's impossible to code review right now.

jpodivin:
> > Determine if we want to be maintaining the log content entirely in-memory or if we can pass a > file-like object into Log Detective as a stream from the zipfile interface. If so, revise the > tempfile to stay available until full processing is complete.
> 
> This is interesting.
> 
> It's true that some logs can be colossal and could crash our service when loaded all in memory. I'm curious what would happen in openshift for such a case.

What's the largest log that can, under reasonable conditions, come out of rpm build? I can't imagine it getting past 1GB.

sgallagher:
> What's the largest log that can, under reasonable conditions, come out of rpm build? I can't imagine it getting past 1GB.

The largest example I can find recently would be for [gcc](https://kojipkgs.fedoraproject.org//packages/gcc/15.0.1/0.10.fc43/data/logs/x86_64/build.log), which is around 135MiB of build.log. That's probably about the maximum size under "reasonable conditions". But it occurs to me that it's also not impossible for a log to end up filled with spew (such as getting caught in a tight loop that includes a log message). I see us as having two options in that case: 1) check the content-length and just refuse to process logs over a certain size or 2) pass the logs as a stream so that Log Detective can handle chunks of it at a time.

If Log Detective needs the whole thing to be in-memory, that probably limits us to option 1.

TomasTomecek:
the content length header is an excellent solution!

TomasTomecek:
Please rebase since the ruff changes are merged now.

logdetective/server/server.py
```diff
@@ -260,3 +273,167 @@ async def analyze_log_stream(build_log: BuildLog):
...
+@app.post("/webhook/gitlab/job_events")
+async def receive_gitlab_job_event_webhook(
+    job_hook: JobHook, background_tasks: BackgroundTasks
+):
+    """Webhook endpoint for receiving job_events notifications from GitLab
+    https://docs.gitlab.com/user/project/integrations/webhook_events/#job-events
+    lists the full specification for the messages sent for job events."""
+
+    # Handle the message in the background so we can return 200 immediately
+    background_tasks.add_task(process_gitlab_job_event, job_hook)
```
TomasTomecek, line 287: still fascinating to me this background tasks functionality is built into a web framework

I hope we'll be able to decode/monitor based on logs what's happening inside

logdetective/server/server.py
```diff
@@ -260,3 +273,167 @@ async def analyze_log_stream(build_log: BuildLog):
...
+
+    # Retrieve the build logs from the merge request artifacts and preprocess them
+    preprocessed_log = await retrieve_and_preprocess_koji_logs(job)
+
+    # Submit log to Log Detective and await the results.
+    response = await submit_log_to_llm(preprocessed_log)
+    preprocessed_log.close()
+
+    # Add the Log Detective response as a comment to the merge request
+    await comment_on_mr(merge_request_id, response)
```
TomasTomecek, line 335: nice and clear :)

logdetective/server/server.py
```diff
@@ -260,3 +273,167 @@ async def analyze_log_stream(build_log: BuildLog):
...
+        # This means: x86_64 > aarch64 > ppc64le > s390x
+        if "x86_64" in failed_arches:
+            failed_arch = "x86_64"
+        elif "aarch64" in failed_arches:
+            failed_arch = "aarch64"
+        elif "ppc64le" in failed_arches:
+            failed_arch = "ppc64le"
+        elif "s390x" in failed_arches:
+            failed_arch = "s390x"
+        else:
```
TomasTomecek, line 415: I wonder how will this code react to SRPM build failures

server/config.yml
```diff
@@ -8,3 +8,6 @@ extractor:
 gitlab:
   url: https://gitlab.com
   api_token: glpat-XXXXXX
+general:
+  packages:
+    - osci-internal-test-package
```
TomasTomecek, line 13: perfect, there is a package we can play with \o/

TomasTomecek (approved):
Love all the comments in the code and how readable it is, very nice work!!

logdetective/server/server.py
sgallagher: I didn't find an example to test, so I probably need to create one...

I'm pretty sure that it still returns an archful log though. Even the noarch builds for python packages, et. al. do that for binary RPMs.

sgallagher:
> the content length header is an excellent solution!

It looks like we actually can stream-process it, which is far more memory-efficient, but I think I'll still implement the header check and limit us to a (configurable) 300 MiB log file. If it's bigger than that, odds are good that the log isn't actually all that useful.


> Please rebase since the ruff changes are merged now.

Done

sgallagher:
I added comments indicating that some of the functions are placeholders that will be implemented in future merge requests. I also added a check for the size of `artifacts.zip` so we will just raise an error if we are being asked to handle very large files. The maximum size is configurable.

logdetective/server/server.py
TomasTomecek: Excellent point, I'll create a task for this.

logdetective/server/server.py
TomasTomecek: Interesting, every buildsystem behaves differently.

If it doesn't work, we'll find out soon enough :)

TomasTomecek (approved):
Amazing! Let's deploy and test it out :)

PR #139, opened 2025-03-20
Implementing interface with /chat/completions endpoint

jpodivin:
* New response model for Log Detective API
* Configuration option for choosing alternative endpoints
* Removed dependence on llama-cpp response models
* Disabled R0913,R0917 pylint rules in several key places, as I consider passing all of these in struct an overkill

### DIFF
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -1,5 +1,5 @@
 from logging import BASIC_FORMAT
-from typing import List, Dict, Optional
+from typing import List, Dict, Optional, Literal
 from pydantic import BaseModel, Field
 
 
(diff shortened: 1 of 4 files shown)


TomasTomecek (approved):
Very nice work! The code looks cleaner and more readable with the new endpoint.

logdetective/server/models.py
```diff
@@ -37,24 +37,29 @@ class JobHook(BaseModel):
     project_id: int
 
 
+class Explanation(BaseModel):
+    """Model of snippet or general log explanation from Log Detective"""
+
+    text: str
+    logprobs: Optional[List[Dict]] = None
+
+
```
sgallagher, line 46: Can we simplify the API for the snippets below by making Explanation behave like a `str` when needed?
```suggestion
class Explanation(BaseModel):
    """Model of snippet or general log explanation from Log Detective"""

    text: str
    logprobs: Optional[List[Dict]] = None

    def __str__(self):
        return self.text
```

logdetective/server/models.py
```diff
@@ -63,14 +68,17 @@ class StagedResponse(Response):
         'line_number': '<location_in_log>' }
     """
 
-    snippets: List[Dict[str, str | Dict | int]]
+    snippets: List[Dict[str, str | Explanation]]
```
sgallagher, line 71:
```suggestion
    snippets: List[Dict[str, Explanation]]
```

sgallagher: Also, the comment above this explaining the member doesn't match the implementation.

logdetective/server/models.py
jpodivin: I get what you mean. I'm not sure how useful it will be in practice but I'll include it. The whole response will have to be redone.

sgallagher: It just seems like it'll be easier on the consumer, rather than them needing to interrogate the returned object to see which type they received. We always return an `Explanation` and we can coerce it into a `str` if an existing client (are there any?) needs it to behave like a string.

sgallagher (approved):
LGTM :+1:

PR #141, opened 2025-03-21
prod deployment: point to latest version

TomasTomecek:
In 78e1db0c27381aaa9f48955870eda21ea4896f20 we have changed the way we
tag our release container images - we no longer have `latest-cuda`.

There we need to pin to specific version for the production deployment.

image

### DIFF
--- a/docker-compose-prod.yaml
+++ b/docker-compose-prod.yaml
@@ -1,7 +1,7 @@
 version: "3"
 services:
   llama-cpp:
-    image: quay.io/logdetective/runtime:latest-cuda
+    image: quay.io/logdetective/runtime:0.2.14-cuda
     build:
       context: .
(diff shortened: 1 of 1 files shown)


FrostyX:
> we no longer have latest-cuda.

Isn't this a step-back?

jpodivin:
> > we no longer have latest-cuda.
> 
> Isn't this a step-back?

Of a sort yes.  There were issues with the suffix being ignored by the metadata action. I've tried to work around it, but it didn't work. Versioning will work until we figure out some way to make both latest tags work reliably.

TomasTomecek:
I'll create a jira card to fix this.

jpodivin (approved):

TomasTomecek:
Rebased and merging.

(who is touching my branch with merge commits :grin:)

PR #142, opened 2025-03-21
Adding log line number information to the snippet

jpodivin:
Location of representative cluster sample, meaning the first log chunk matching created cluster, is now stored and used as part of prompt when using the Drain extractor.

Further changes:

* Fixing typo in SNIPPET_PROMPT_TEMPLATE
* Ammending StagedResponse model docstring
* Return type annotations for get_chunks utility function
* Refactoring snippet joining into format_analyzed_snippets function

### DIFF
--- a/logdetective/utils.py
+++ b/logdetective/utils.py
@@ -1,12 +1,12 @@
 import logging
 import os
-from typing import Iterator, List, Dict
+from typing import Iterator, List, Dict, Tuple, Generator
 from urllib.parse import urlparse
 import numpy as np
 import requests
 
 from llama_cpp import Llama, CreateCompletionResponse, CreateCompletionStreamResponse
-from logdetective.constants import PROMPT_TEMPLATE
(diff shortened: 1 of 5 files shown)


logdetective/utils.py
```diff
@@ -142,18 +148,41 @@ def retrieve_log_content(log_path: str) -> str:
     return log
 
 
-def format_snippets(snippets: list[str]) -> str:
+def format_snippets(snippets: list[str] | list[Tuple[int, str]]) -> str:
     """Format snippets, giving them separator, id and finally
-    concatenating them.
+    concatenating them. If snippets have line number attached,
+    include that in prompt.
```
TomasTomecek, line 154: very nice! I'm looking fwd testing this out how will LLMs handle those line numbers

TomasTomecek (approved):
Pretty clear changes, thanks!

We need to start introducing test cases for these in the future.

Issue #134, opened 2025-03-17
ModuleNotFoundError: No module named 'pydantic'


xsuchy:
Build for F42 fails 
  https://src.fedoraproject.org/rpms/python-logdetective/pull-request/10
in %check phase with

```
modules -e logdetective.server
Check import: logdetective
Check import: logdetective.constants
Check import: logdetective.extractors
Check import: logdetective.logdetective
Check import: logdetective.server.models
Traceback (most recent call last):
  File "/usr/lib/rpm/redhat/import_all_modules.py", line 171, in <module>
    main()
    ~~~~^^
  File "/usr/lib/rpm/redhat/import_all_modules.py", line 167, in main
    import_modules(modules)
    ~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/lib/rpm/redhat/import_all_modules.py", line 100, in import_modules
    importlib.import_module(module)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 1026, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/builddir/build/BUILD/python-logdetective-0.2.13-build/BUILDROOT/usr/lib/python3.13/site-packages/logdetective/server/models.py", line 2, in <module>
    from pydantic import BaseModel
ModuleNotFoundError: No module named 'pydantic'
```

TomasTomecek:
I'd say the easiest fix would be to not include server code into the release tarball. So we are not polluting the CLI with the server stuff.

Or properly package it.

TomasTomecek:
I'm fine packaging it since it will make testing it via packit and testing farm much easier.

FrostyX:
This should be fixed by https://bodhi.fedoraproject.org/updates/FEDORA-2025-b0e74039c1

PR #145, opened 2025-03-22
Use poetry extras instead of groups

FrostyX:
See #134

From the Poetry documentation:
https://python-poetry.org/docs/managing-dependencies/

> Dependency groups, other than the implicit main group, must only contain
> dependencies you need in your development process. Installing them is only
> possible by using Poetry.
>
> To declare a set of dependencies, which add additional functionality to the
> project during runtime, use extras instead. Extras can be installed by the end
> user using pip.

As a consequence, there is IMHO no way to install them in a RPM spec file, other than duplicating them as `BuildRequires`.

There is a support for installing groups through `%pyproject_buildrequires`, however, to make things more complicated, Poetry groups are not the same as PEP 735 groups. I think.


TomasTomecek:
Ah, that's unfortunate. Thanks for investigating!

So shall we just add those packages in BuildRequires to solve the build problems?

FrostyX:
> So shall we just add those packages in BuildRequires to solve the build problems?

That depends. If we merge this PR and publish a new release, then the spec file can remain untouched. If you don't like this approach and want to keep the previous one, yea, I can easily add `BuildRequires` to the specfile.

Depends on your preference.

FrostyX:
> Extras have been deprecated. Documentation suggests using project.optional-dependencies https://python-poetry.org/docs/pyproject/#optional-dependencies instead.

Would be nice if poetry command could generate them, then :D
Let me try

FrostyX:
I think this will be a Poetry 2.0 thing, which will be F43+. On my F41, poetry doesn't seem to be able to install the dependencies specified like this:

```
[project.optional-dependencies]
server = [ "pydantic>=2.10.6,<3.0", "fastapi>=0.115.11,<1.0", "pyyaml>=6.0.2,<7.0" ]
```

And I didn't have any success with the spec part either:

```
%generate_buildrequires
%pyproject_buildrequires -x server  # Nor when using -g
```

Personally, I'd go for the deprecated extras until we have Poetry 2.0 in the stable Fedora releases.

hroncok:
`[project.optional-dependencies]` is only supported in newer poetry-core. If you want to support both old and new Fedora, have a look at flit-core or hatchling instead. If you insist on using Poetry, you need to use the deprecated `[tool.poetry.extras]` with `{version="...", optional=true}` in `[tool.poetry.dependencies]`.

FrostyX:
Thank you very much for the comment @hroncok.
I think, it corroborates what I was trying to say.

> have a look at flit-core or hatchling instead.

I'll do as the group decides. But my personal vote would be keeping poetry for the time being, since we are severely time constrained right now.

jpodivin:
Ok, it seems pretty conclusive that we should use extras for the time being.

FrostyX:
> Ok, it seems pretty conclusive that we should use extras for the time being.

Or finish the switch to hatch in PR #148

FrostyX:
As agreed on the meeting, we will continue with poetry and delay the hatch migration a bit.

PR #146, opened 2025-03-24
Create a database for storing metrics

majamassarini:
Add first metrics table: store data for received requests and given responses.


Makefile
```diff
@@ -0,0 +1,40 @@
...
+server-up:
+	$(COMPOSE_ENGINE) up --build --force-recreate -d server
+
+# run alembic revision in another pod
+alembic-generate-revision: server-up
+	@echo "Waiting for server to start and invoking alembic upgrade head..."
+	sleep 3
+
+	@echo "Checking if database is ready..."
+	$(CONTAINER_ENGINE) run --rm --network logdetective_default \
```
TomasTomecek, line 23: I think `$(COMPOSE_ENGINE) exec postgres` should serve this purpose better

Makefile
```diff
@@ -0,0 +1,40 @@
...
+		-e POSTGRESQL_HOST=postgres \
+		-e POSTGRESQL_DATABASE=$(POSTGRESQL_DATABASE) \
+		-v $(PWD)/alembic:/src/alembic:rw,z \
+		-v $(PWD)/alembic.ini:/src/alembic.ini:ro,z \
+		--network logdetective_default \
+		localhost/logdetective/runtime:latest \
+		bash -c "cd /src && alembic revision -m \"$(CHANGE)\" --autogenerate"
+
+	@echo "Shutting down server..."
+	$(COMPOSE_ENGINE) down
```
TomasTomecek, line 40: this is kind of unexpected to me

TomasTomecek (commented):
Thank you Maja, amazing work!!

I'll continue with the review tomorrow & try it locally.

Makefile
```diff
@@ -0,0 +1,40 @@
+include .env
+
+CONTAINER_ENGINE ?= $(shell command -v podman 2> /dev/null || echo docker)
+COMPOSE_ENGINE ?= $(shell command -v podman-compose 2> /dev/null || echo docker-compose)
+MY_ID ?= $(shell id -u)
+CHANGE ?= New migration
+
+.PHONY: server
```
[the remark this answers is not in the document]
majamassarini: Thanks for catching it, I always forgot to update .PHONY  😅

Makefile
majamassarini: I am not sure, in that way we will not check the "network connection", right? But if you think this is too complex I can also drop it. It only a bit useful to give a more meaningful error message in this case...

Makefile
majamassarini: The prerequisite of this makefile's target is `podman-compose us server` , because it is the server starting file that calls `alembic upgrade head`, and here I am cleaning things. Probably I should improve it with `podman-compose down server`. Is this what you meant?

Makefile
TomasTomecek: if your proposed solution works, let's keep it :) I still didn't have time try it locally

sorry I did not explain what I had in mind precisely, I'll send followup PR once this is merged, that should make the explanation easier

Makefile
TomasTomecek: kind of yes

my point was: if nothing is running and I want to just create a migration, then this is 100% expected - stand up the server, create migration, bring it down

but if it's already running, then it will be brought down which would be unexpected to me, hence I would comment this section out if I wanna keep the server up

this is definitely a nit pick, I'm not sure which of those 2 cases will be more common

if you propose this and feel like this is the right thing to do, let's keep it; I just wanted to be sure we understand each other's thinking process

logdetective/server/database/base.py
```diff
@@ -0,0 +1,63 @@
...
+def init():
+    """Init db"""
+    Base.metadata.create_all(engine)
+    logger.debug("Database initialized")
+
+
+def clean():
+    """Clean db"""
+    Base.metadata.drop_all(engine)
+    logger.debug("Database cleaned")
```
FrostyX, line 63: Nitpick: This is a very dangerous operation, destroying the whole database. Can we please rename the function so that nobody accidentally calls it? I would go with something like `delete` or `destroy`.

logdetective/server/database/model.py
```diff
@@ -0,0 +1,85 @@
...
+
+class AnalyzeRequestMetrics(Base):
+    """Store data related to received requests and given responses"""
+    __tablename__ = "analyze_request_metrics"
+
+    class EndpointType(enum.Enum):
+        """Different analyze endpoints"""
+        ANALYZE = "analyze_log"
+        ANALYZE_STAGED = "analyze_log_staged"
+        ANALYZE_STREAM = "analyze_log_stream"
```
FrostyX, line 18: Nitpick: I dislike nesting classes and functions when possible. Can we please move this to top-level?

logdetective/server/database/model.py
```diff
@@ -0,0 +1,85 @@
...
+        response_certainty: float,
+    ) -> None:
+        """Update an AnalyzeRequestMetrics line
+        with data related to the given response"""
+        with transaction(commit=True) as session:
+            metrics = session.query(AnalyzeRequestMetrics).filter_by(id=id_).first()
+            metrics.response_sent_at = response_sent_at
+            metrics.response_length = response_length
+            metrics.response_certainty = response_certainty
+            session.add(metrics)
```
FrostyX, line 85: I am not sure if I like the `create` and `upate` methods as a part of the model classes. It seems to add a lot of boilerplate and one more hoop to jump through, while I don't really see any benefit. Personally, I would move the relevant part of the code directly to `add_new_metrics` and `update_metrics`.

logdetective/server/metric.py
```diff
@@ -0,0 +1,62 @@
...
+                if "text" in choice
+            )
+        response_certainty = (
+            response.response_certainty
+            if hasattr(response, "response_certainty")
+            else None
+        )
+        database.model.AnalyzeRequestMetrics.update(
+            metrics_id, response_sent_at, response_length, response_certainty
+        )
```
FrostyX, line 41: I think the `add_new_metrics` and `update_metrics` functions can be top-level and potentially called without the decorator if needed?

FrostyX (commented):
Wow, this was an impressive amount of work, thank you @majamassarini 

I left some comments.

Edit: I forgot to say - It works great. I started the containers, sent some request, it appeared in the `analyze_request_metrics` table.

.pylintrc.tests
```diff
@@ -0,0 +1,15 @@
...
+# Disable specific messages for tests and alembic code
+disable=
+    redefined-outer-name,  # Allow redefining names from outer scope (common in pytest fixtures)
+    invalid-name,          # Allow non-standard naming in tests, we may want to have name like mock_AnalyzeRequestMetrics
+    missing-docstring,     # Tests often don't need docstrings
+    protected-access,      # Tests often need to access protected members
+    too-many-arguments,    # Test functions often need many arguments
+    too-many-locals,       # Tests may use many local variables
+    unused-argument,       # Fixtures may be unused in some tests
+    no-member              # alembic code often raise this warning as in Module 'alembic.op' has no 'f' member
```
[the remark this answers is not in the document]
majamassarini: I added the reformatting tool to the pre-commit hook in this PR initially, but removed it when you mentioned we don't want a code formatter.

After thinking about it more, I honestly agree with Jiri. I believe the pre-commit hook formatter is our best option. It will format only the newly staged code, whereas currently I'm manually applying reformatting to all files while trying to skip code I didn't write. This works for now, but I'm not confident it's sustainable in the long run.

docker-compose-prod.yaml
```diff
@@ -22,17 +22,32 @@ services:
...
-    # --no-reload: doesn't work in a container - `PermissionError: Permission denied (os error 13) about ["/proc"]`
-    # command: fastapi run /src/logdetective/server.py --host 0.0.0.0 --port $LOGDETECTIVE_SERVER_PORT --no-reload
-    command: ["gunicorn", "-c", "/src/server/gunicorn-prod.config.py", "logdetective.server.server:app"]
+    entrypoint: ["/run_server.sh"]
+
+  postgres:
+    container_name: postgres
+    image: quay.io/sclorg/postgresql-15-c9s
+    volumes:
+      - database_data:/var/lib/postgresql/data
```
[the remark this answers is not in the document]
majamassarini: I don't have any preference. I'll let you choose since you've already deployed it in production and you understand our needs there better than I do.

logdetective/server/database/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+__all__ = [
+    "get_pg_url",
+    "Base",
+    "engine",
+    "SessionFactory",
+    "transaction",
+    "init",
+    "clean",
+    "AnalyzeRequestMetrics",
+]
```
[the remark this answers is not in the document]
majamassarini: I think this is mainly usefull for the linter, if I don't add the __all__ section, my above imports will raise an unused-import warning.
I think it is a good practice to state which is the "public interface" of a module.

logdetective/server/database/base.py
majamassarini: Just a note on why I introduced it - I'm using it in the unit tests.

logdetective/server/metric.py
majamassarini: Yes, I agree. I will move them up.

logdetective/server/database/__init__.py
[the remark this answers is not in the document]
jpodivin: You don't really have to import them at all here. And we don't really have anything that would be called "private" in that module. Even if we had that sort of functions there, the convention is to simply prefix them with `_` to indicate that.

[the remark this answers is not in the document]
majamassarini: As I wrote I like using namespaces 🙂
for being able to call `database.init()` I need this import. I removed the clean/destroy from the list, so we can not call database.destroy(). I think this quite explicit on what we want our database API to be.

You can always do a `from logdetective.server.database.base import destroy` (and now I am doing it from the unittests) but it is way less explicit.

But if I am the only one that likes namespaces I can drop the imports 🙂

.pylintrc.tests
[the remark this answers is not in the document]
jpodivin: The end of file specifically is already part of the hooks we have https://github.com/fedora-copr/logdetective/blob/main/.pre-commit-config.yaml#L8

logdetective/server/database/__init__.py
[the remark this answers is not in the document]
jpodivin: Yeah, but being explicit is a good thing. It makes following code much easier for others.

.pylintrc.tests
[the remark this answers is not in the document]
majamassarini: Oh ok, I didn't realize the pre-commit hook is not called anymore then 😅

logdetective/server/database/model.py
majamassarini: Since I removed the definition of the database namespace I thought there is no reason to keep this. So I have moved the enum definition up.

Containerfile
```diff
@@ -26,3 +29,7 @@ RUN pip3 install llama_cpp_python==0.2.85 sse-starlette starlette-context \
 #     && huggingface-cli download TheBloke/Mistral-7B-Instruct-v0.2-GGUF mistral-7b-instruct-v0.2.Q4_K_M.gguf --local-dir /models --local-dir-use-symlinks False
 
 COPY ./logdetective/ /src/logdetective/
+COPY ./alembic.ini /src/alembic.ini
+COPY ./alembic /src/alembic
+
+WORKDIR /src
```
FrostyX, line 35: Still missing EOF here

Makefile
```diff
@@ -0,0 +1,42 @@
...
+		-e POSTGRESQL_HOST=postgres \
+		-e POSTGRESQL_DATABASE=$(POSTGRESQL_DATABASE) \
+		-v $(PWD)/alembic:/src/alembic:rw,z \
+		-v $(PWD)/alembic.ini:/src/alembic.ini:ro,z \
+		--network logdetective_default \
+		localhost/logdetective/runtime:latest \
+		bash -c "cd /src && alembic revision -m \"$(CHANGE)\" --autogenerate"
+
+	@echo "WARNING: Shutting down server..."
+	$(COMPOSE_ENGINE) down server
```
FrostyX, line 42: Also here and in other files as well

FrostyX (commented):
Still missing EOF in some files, but after that, we can IMHO merge.
I would prefer rebasing and squashing some commits together.

majamassarini:
@FrostyX I squashed some of the commits and manually run the pre-commit hook in every other commit. It seems to me that now the formatting is ok.

TomasTomecek (approved):
Just tried it locally, very nice work!! All smooth.

FrostyX (approved):
Lets gooo :rocket:

PR #147, opened 2025-03-24
New latest

jpodivin:
Some aspects of the action were not properly documented. 
Tag `latest` is given special treatment, which takes precedence over other rules.

The order of actions also seems to matter, as some environment vars stick around. 

Tested in https://github.com/jpodivin/logdetective/actions/runs/14032242732/job/39282235566

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,4 +1,4 @@
-name: Docker Build and Publish
+name: Container Image Build and Publish
 
 # This workflow uses actions that are not certified by GitHub.
 # They are provided by a third-party and are governed by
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
thank you!

FrostyX:
I released a new version with this change, @jpodivin can you please take a look if this had the desired effect? I am afraid it didn't but I am not sure if I understand the change correctly. Anyway, looking here

https://quay.io/repository/logdetective/runtime?tab=tags

the following tags got updated/created:

- latest
- sha-8b4b8d0-cuda
- 0.3.3-cuda
- sha-8b4b8d0
- 0.3.3

but latest-cuda didn't.

PR #148, opened 2025-03-24
Convert to hatch

sgallagher:
If we wanted to switch to hatch[ling], this could be a start for that.

Note: this is quick-and-dirty and minimally-tested.


TomasTomecek (commented):
+1 to switch to hatch if it makes our packaging work easier

pyproject.toml
```diff
@@ -20,32 +17,38 @@ classifiers = [
...
+    "pydantic>=2.10.6, < 3.0.0",
+    "pyyaml>=6.0.2, <7.0.0",
+]
+
+[tool.hatch.envs.hatch-build]
+dependencies = [
+  "cython",
+]
+
+[projectscripts]
```
FrostyX, line 51: You are missing a dot here, should be `[project.scripts]`

pyproject.toml
```diff
@@ -20,32 +17,38 @@ classifiers = [
...
+
+dependencies = [
+    "requests>0.2.31",
+    "llama-cpp-python>0.2.56, !=0.2.86",
+    "drain3 >=0.9.11, < 1.0.0",
+    "huggingface-hub>0.23.2",
+    # rawhide has numpy 2, F40 and F41 are still on 1.26
+    # we need to support both versions
+    "numpy>=1.26.0",
+    "python-gitlab >=5.6.0, < 6.0.0",
```
FrostyX, line 30: In order to build for F41, we need to relax the dependency to `"python-gitlab >=4.10.0, < 6.0.0"`

pyproject.toml
```diff
@@ -20,32 +17,38 @@ classifiers = [
...
-[build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
 
-[tool.poetry.scripts]
+[project.optional-dependencies]
+server = [
+    "fastapi>=0.115.8, < 1.0.0",
+    "pydantic>=2.10.6, < 3.0.0",
+    "pyyaml>=6.0.2, <7.0.0",
```
FrostyX, line 43: In order to build for F41, we need to relax the dependency to `"pyyaml>=6.0.1, <7.0.0"`

FrostyX (commented):
Thank you for the change @sgallagher,
I tried to build an RPM package based around this change, and it worked fine (apart from a few minor hiccups that I am pointing out in my review)

The only reason why I wanted to stay with Poetry was that we already had it, so it was easier to keep it than getting rid of it. Since you already proposed this PR, it is +1 from me.

pyproject.toml
sgallagher: Whoops! Thanks for catching that.

sgallagher:
> Two things to point out.
> 
>     * I only tried building the package, not actually using it.

So, no different from anything that Google ships, then? (I'll see myself out)

>     * Please see `git grep poetry`. We also need to change all of those

Yeah, I'll take a look.

PR #149, opened 2025-03-24
Start logging to a file

FrostyX:
Fix LD-28

### DIFF
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -1,3 +1,4 @@
+from logging import BASIC_FORMAT
 from typing import List, Dict, Optional
 from pydantic import BaseModel, Field
 
@@ -118,6 +119,25 @@ class GitLabConfig(BaseModel):
         self.max_artifact_size = int(data.get("max_artifact_size")) * 1024 * 1024
 
 
+class LogConfig(BaseModel):
+    """Logging configuration"""
+
+    name: str = "logdetective"
+    level: str | int = "INFO"
+    path: str | None = None
+    format: str = BASIC_FORMAT
+
+    def __init__(self, data: Optional[dict] = None):
+        super().__init__()
+        if data is None:
+            return
+
+        self.name = data.get("name", "logdetective")
+        self.level = data.get("level", "INFO").upper()
+        self.path = data.get("path")
+        self.format = data.get("format", BASIC_FORMAT)
+
+
 class GeneralConfig(BaseModel):
     """General config options for Log Detective"""
 
@@ -134,6 +154,7 @@ class GeneralConfig(BaseModel):
 class Config(BaseModel):
     """Model for configuration of logdetective server."""
 
+    log: LogConfig = LogConfig()
     inference: InferenceConfig = InferenceConfig()
     extractor: ExtractorConfig = ExtractorConfig()
     gitlab: GitLabConfig = GitLabConfig()
@@ -145,6 +166,7 @@ class Config(BaseModel):
         if data is None:
             return
 
+        self.log = LogConfig(data.get("log"))
         self.inference = InferenceConfig(data.get("inference"))
         self.extractor = ExtractorConfig(data.get("extractor"))
         self.gitlab = GitLabConfig(data.get("gitlab"))
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -1,6 +1,5 @@
 import asyncio
 import json
-import logging
 import os
 import re
 import zipfile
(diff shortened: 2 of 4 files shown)


logdetective/server/server.py
```diff
@@ -48,6 +47,18 @@
 FAILURE_LOG_REGEX = re.compile(r"(\w*\.log)")
 
 
+if SERVER_CONFIG.log.path:
+    logging.basicConfig(
```
sgallagher, line 51: Using `logging.basicConfig()` means that we're setting the log configuration for ANYTHING that uses the python logging facility. That includes things like `requests` and `urllib3` at the low level. This can get a little noisy.

Can we restrict the logging setup to just the logger we're going to use?

logdetective/server/models.py
```diff
@@ -118,6 +119,37 @@ def __init__(self, data: Optional[dict] = None):
...
 
+LogLevel = Literal[
+    "CRITICAL",
+    "FATAL",
+    "ERROR",
+    "WARN",
+    "WARNING",
+    "INFO",
+    "DEBUG",
+    "NOTSET",
```
sgallagher, line 130: Is there a reason to redefine these rather than just use `logging.CRITICAL`, `logging.DEBUG`, etc?

When setting loglevel (either with `basicConfig()` or `logger.setlevel()`) we can just pass `levelstring.upper()` and it will be interpreted properly. What does this redefinition gain us?

FrostyX: The reason was to have a specific type here:

```python
class LogConfig(BaseModel):
    ...
    level: LogLevel = "INFO"
```

But if you want, we can drop the list and use:
```python
class LogConfig(BaseModel):
    ...
    level: str | int = "INFO"
```

logdetective/server/server.py
FrostyX: Good point, thank you.

FrostyX:
Updated, PTAL

sgallagher (approved):
LGTM :+1:

server/config.yml
```diff
@@ -1,3 +1,6 @@
+log:
+  level: "INFO"
+  path: "log/logdetective.log"
```
TomasTomecek, line 3: I'd argue to default to `/dev/stderr` given we are in the container world

What does everyone think?

TomasTomecek (commented):
Very nice work! But since we run in containers, I could not see a handler for stderr. To me it makes sense to log in both, actual file and stderr, but I personally value stderr the most.

sgallagher (changes requested):
Actually, one more thing: could you make `get_log()` into a singleton pattern?

I'd like to be able to call `LOG = get_log(SERVER_CONFIG)` from multiple python files and just have it return the log without reinitializing it.

server/config.yml
sgallagher: It's actually already logging to stdout and the logfile. See [utils.py](https://github.com/fedora-copr/logdetective/blob/043e108430a68ad82d558d0fde3801c24d3a0f7f/logdetective/server/utils.py#L29).

logdetective/server/utils.py
```diff
@@ -13,3 +14,26 @@ def load_server_config(path: str | None) -> Config:
...
+    stream_handler = logging.StreamHandler()
+    stream_handler.setFormatter(logging.Formatter(config.log.format))
+    log.addHandler(stream_handler)
+
+    # Log to file
+    if config.log.path:
+        file_handler = logging.FileHandler(config.log.path)
+        file_handler.setFormatter(logging.Formatter(config.log.format))
+        log.addHandler(file_handler)
+    return log
```
sgallagher, line 39:
```suggestion
def get_log(config: Config):
    """
    Initialize a logger for this server
    """
    log = logging.getLogger(config.log.name)
    if config.log.initialized:
        return log
    
    log.setLevel(config.log.level)

    # Drop the default handler, we will create it ourselves
    log.handlers = []

    # STDOUT
    stream_handler = logging.StreamHandler()
    stream_handler.setFormatter(logging.Formatter(config.log.format))
    log.addHandler(stream_handler)

    # Log to file
    if config.log.path:
        file_handler = logging.FileHandler(config.log.path)
        file_handler.setFormatter(logging.Formatter(config.log.format))
        log.addHandler(file_handler)

    config.log.initialized = True
    return log
```

logdetective/server/models.py
```diff
@@ -118,6 +119,25 @@ def __init__(self, data: Optional[dict] = None):
         self.max_artifact_size = int(data.get("max_artifact_size")) * 1024 * 1024
 
 
+class LogConfig(BaseModel):
+    """Logging configuration"""
+
+    name: str = "logdetective"
+    level: str | int = "INFO"
+    path: str | None = None
+    format: str = BASIC_FORMAT
```
sgallagher, line 128:
```suggestion
    name: str = "logdetective"
    level: str | int = "INFO"
    path: str | None = None
    format: str = BASIC_FORMAT
    initialized: bool = False
```

sgallagher (changes requested):
Adding a suggestion for a singleton pattern for `get_log()`.

jpodivin:
+1 on @sgallagher  recommendations

FrostyX:
I used the `.upper()` suggestion but I did the singleton pattern slightly differently without abusing `LogConfig`.

FrostyX:
I am merging so it gets to the new release. We can improve in a follow-up PR.

server/config.yml
TomasTomecek: oh, so it's extra; got confused, thanks for explanation!

PR #150, opened 2025-03-25
Running tox linters in CI

jpodivin:


majamassarini (approved):
Thanks! For me this is an improvement :) But let's wait what others think.

PR #154, opened 2025-03-26
Backporting changes from production

FrostyX:
We have these uncommitted changes on logdetective01 server.


TomasTomecek (approved):
Thank you for updating the containers to 41 as well!

Glad the python 3.12 -> 3.13 change went well

PR #155, opened 2025-03-27
Relax dependencies for F40 and F41

FrostyX:
With these, I am able to successfully build RPM package for all currently active Fedora versions.


TomasTomecek (approved):
LGTM

PR #156, opened 2025-03-27
Make inference URL and token configurable via config file

TomasTomecek:
LD-32

This PR moves inference server configuration to the server config file and removes related values from .env.


docker-compose-prod.yaml
```diff
@@ -5,7 +5,7 @@ services:
     build:
       context: .
       dockerfile: ./Containerfile.cuda
-    hostname: "${LLAMA_CPP_HOST}"
+    hostname: llama-cpp-server
```
jpodivin, line 8: Please add note that this has to match configuration in the config file. It is obvious, but previously we didn't have to care about that.

TomasTomecek: will add it

TomasTomecek:
Yes, I realized I should have merged this last week but I also didn't wanna force others to rebase.

FrostyX (approved):

PR #159, opened 2025-03-28
Minimize differences between devel and production composes

FrostyX:
My goal is to eventually remove the `docker-compose-prod.yaml` completely and keep just `docker-compose.yaml` but this is IMHO a good first step.

I tested this in production and it works fine.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,7 +1,7 @@
 version: "3"
 services:
   llama-cpp:
-    image: logdetective/runtime:latest-cuda
+    image: quay.io/logdetective/runtime:0.3.3-cuda
     build:
       context: .
       dockerfile: ./Containerfile.cuda
@@ -20,7 +20,7 @@ services:
     # security_opt:
     #   - "label=disable"
   server:
-    image: logdetective/runtime:latest
+    image: quay.io/logdetective/runtime:0.3.3
     depends_on:
       - postgres
     build:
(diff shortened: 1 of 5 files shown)


FrostyX:
Rebased, PTAL.

files/run_server.sh
```diff
@@ -19,8 +19,11 @@ if [[ $n -eq $ATTEMPTS ]]; then
...
+else
+    conf="/src/server/gunicorn-dev.config.py"
+fi
+
 echo "Starting application server..."
-# --no-reload: doesn't work in a container - `PermissionError: Permission denied (os error 13) about ["/proc"]`
-# command: fastapi dev /src/logdetective/server.py --host 0.0.0.0 --port $LOGDETECTIVE_SERVER_PORT --no-reload
-# timeout set to 240 - 4 minutes should be enough for one LLM execution locally on a CPU
-gunicorn -k uvicorn.workers.UvicornWorker --timeout 240 logdetective.server.server:app -b 0.0.0.0:${LOGDETECTIVE_SERVER_PORT}
+gunicorn -c "$conf" logdetective.server.server:app
```
TomasTomecek, line 29:
```suggestion
exec gunicorn -c "$conf" logdetective.server.server:app
```
so that gunicorn process replaces the bash process

TomasTomecek (approved):
Well done!

docker-compose.yaml
```diff
@@ -1,7 +1,6 @@
-version: "3"
 services:
   llama-cpp:
-    image: logdetective/runtime:latest-cuda
+    image: quay.io/logdetective/runtime:0.3.3-cuda
```
sgallagher, line 3: We're going to need to keep this in sync with every release; should we update the release scripting to do that?

FrostyX: I dislike it too, but this is only a temporary measure. My next JIRA ticket is to fix the action that pushes the images. So once that is done, we will go back to `latest`.

PR #160, opened 2025-03-28
From pypi install psycopg2-binary instead of psycopg2

majamassarini:
Otherwise pypi may try to build the wheel.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,11 +43,13 @@ pydantic = {version = "^2.8.2", optional = true }
 fastapi = {version = ">=0.111.1", optional = true }
 pyyaml = {version = "^6.0.1", optional = true }
 sqlalchemy = {version = "^2.0.36", optional = true }
+psycopg2-binary = {version = "^2.9.9", optional = true }
 psycopg2 = {version = "^2.9.9", optional = true }
 alembic = {version = "^1.13.3", optional = true }
 
 [tool.poetry.extras]
 server = ["pydantic", "fastapi", "pyyaml", "sqlalchemy", "psycopg2", "alembic"]
+server-testing = ["pydantic", "fastapi", "pyyaml", "sqlalchemy", "psycopg2-binary", "alembic"]
 
 [build-system]
 requires = ["poetry-core"]
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,13 +2,13 @@
 
 [[package]]
 name = "alembic"
-version = "1.15.1"
+version = "1.15.2"
 description = "A database migration tool for SQLAlchemy."
 optional = true
 python-versions = ">=3.9"
 files = [
-    {file = "alembic-1.15.1-py3-none-any.whl", hash = "sha256:197de710da4b3e91cf66a826a5b31b5d59a127ab41bd0fc42863e2902ce2bbbe"},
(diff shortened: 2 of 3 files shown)


FrostyX:
This fails to build in Fedora :-/

```
Problem: nothing provides requested (python3dist(psycopg2-binary) < 3~~ with python3dist(psycopg2-binary) >= 2.9.9)
```

majamassarini:
I quote the `psycopg2` pypi page:

---
You can also obtain a stand-alone package, not requiring a compiler or external libraries, by installing the [psycopg2-binary](https://pypi.org/project/psycopg2-binary/) package from PyPI:
```
$ pip install psycopg2-binary
```
The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

---

We are using pypi packages for upstream testing purposes, so we are supposed to use there the binary package.

The problem of using `psycopg2-binary` is that `pyproject_buildrequires` is not able to solve the rpm package name, because, obviously, it does not exist a `python-psycopg2-binary` package.

The problem of using `psycopg2` is that we should keep track, in our dependencies, of all the dependencies required to build psycopg2 itself (even though as `dev` dependencies I don't think we should, they are not our dependencies).


I spoke with @nforro, he suggested me to use `sed` in downstream to change the `.toml` file before processing the requirements specfile part, like he does  [here](https://src.fedoraproject.org/rpms/awscli2/blob/rawhide/f/awscli2.spec#_64). 
I agree with him that this sounds the better approach. 
So if we are all wanting to follow this way I will insert the `sed` line in the downstream specfile and then I can rerun checks here.

FrostyX:
Thank you for the investigation @majamassarini.

> I spoke with @nforro, he suggested me to use sed in downstream to change the .toml file before processing the requirements specfile part, like he does [here](https://src.fedoraproject.org/rpms/awscli2/blob/rawhide/f/awscli2.spec#_64).

I can do that :-)

jpodivin:
Yes. Go for it.

majamassarini:
Tomorrow I will merge and test this [downstream PR](https://src.fedoraproject.org/rpms/python-logdetective/pull-request/13). Let's me know if I have to change it, thanks!

pyproject.toml
```diff
@@ -40,11 +40,11 @@ pydantic = {version = "^2.8.2", optional = true }
 fastapi = {version = ">=0.111.1", optional = true }
 pyyaml = {version = "^6.0.1", optional = true }
 sqlalchemy = {version = "^2.0.36", optional = true }
-psycopg2 = {version = "^2.9.9", optional = true }
+psycopg2-binary = {version = "^2.9.9", optional = true }
 alembic = {version = "^1.13.3", optional = true }
 
 [tool.poetry.extras]
-server = ["pydantic", "fastapi", "pyyaml", "sqlalchemy", "psycopg2", "alembic"]
+server = ["pydantic", "fastapi", "pyyaml", "sqlalchemy", "psycopg2-binary", "alembic"]
```
nforro, line 47: Since you are already using extras, maybe you could define another one, called let's say `server-source`, that would include `psycopg2` instead of `psycopg2-binary`, and you could use that one downstream:
```specfile
%generate_buildrequires
%pyproject_buildrequires -x server-source
```

majamassarini: Thanks a lot @nforro I think this is another great idea, better than having a `sed` expression downstream. I try it now!

nforro (approved):
Ok, this makes more sense :slightly_smiling_face:

jpodivin (approved):

PR #161, opened 2025-03-28
Webhook: submit build failure to LLM and log the response

sgallagher:
## webhook: Submit the log text to the LLM

This will perform staged analysis. As a temporary placeholder, it also
logs the response in human-readable format. This will be replaced by a
jinja2 template in a future patch.

Fixes: https://issues.redhat.com/browse/LD-16

## Make /analyze/staged body a separate function

This will allow it to be reused by the webhook handler without needing a
URL for the log text.

Depends-on: #139 
This takes advantage of the new, much more understandable `Explanation` model.

### DIFF
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -21,7 +21,7 @@ class JobHook(BaseModel):
 
     # The identifier of the job. We only care about 'build_rpm' and
     # 'build_centos_stream_rpm' jobs.
-    build_name: str = Field(pattern=r"^build(_.*)?_rpm$")
+    build_name: str = Field(pattern=r"^build.*rpm$")
 
     # A string representing the job status. We only care about 'failed' jobs.
     build_status: str = Field(pattern=r"^failed$")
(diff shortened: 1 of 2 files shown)


logdetective/server/models.py
```diff
@@ -37,40 +37,59 @@ class JobHook(BaseModel):
...
-    snippets:
-        list of dictionaries {
-        'snippet' : '<original_text>,
-        'comment': CreateCompletionResponse,
-        'line_number': '<location_in_log>' }
+    snippets: list of AnalyzedSnippet objects
     """
 
-    snippets: List[Dict[str, str | Dict | int]]
+    snippets: List[AnalyzedSnippet]
```
TomasTomecek, line 82: This is lovely that it got its own model

TomasTomecek (approved):
Well done, love the decomposition of the functions.

logdetective/server/models.py
sgallagher: Just so we're clear, this is coming from #139 
I rebased this set of patches atop those because it vastly simplifies what I needed to do here.

sgallagher:
Retiring in favor of #164

PR #163, opened 2025-03-31
Generate Gitlab merge request comments

sgallagher:
This patch adds support for converting the results of the staged
analysis into a comment to be submitted against the original merge
request that triggered the analysis.

It takes advantage of Jinja 2 templating to generate the content in
Markdown format.

This patch also modifies the return values of the log preprocessing
function to return the URL of the detected log so that it can be linked
in the comment.

Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

It depends on (and therefore includes in this stack of patches) #161

PR #164, opened 2025-04-01
Complete the initial Gitlab Pipeline integration

sgallagher:
This merge request now includes and supersedes both #161 and #163 

After this patch-set, the Gitlab integration will meet its initial MVP (minimum viable project) status by:

* Receiving notification from the pipeline of failures in CentOS Stream or RHEL RPM builds in the RHEL-on-Gitlab pipeline.
* Will interrogate the contents of the Koji build logs to determine the most likely log to contain the failure information.
* Will then submit that log to the LLM for analysis using the "staged" mechanism.
* Will retrieve the resulting analysis, both general and with individual snippets and construct a Markdown comment.
* Will attach that markdown comment as a new discussion thread on the merge request that spawned the failed pipeline.

Fixes: https://issues.redhat.com/browse/LD-16
Fixes: https://issues.redhat.com/browse/LD-17

### DIFF
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -318,6 +318,12 @@ async def analyze_log_staged(build_log: BuildLog):
     while lacking  result, params or query fields.
     """
     log_text = process_url(build_log.url)
+
+    return await perform_staged_analysis(log_text=log_text)
+
+
+async def perform_staged_analysis(log_text: str) -> StagedResponse:
+    """Submit the log file snippets to the LLM and retrieve their results"""
     log_summary = mine_logs(log_text)
 
     # Process snippets asynchronously
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -451,11 +451,12 @@ async def process_gitlab_job_event(job_hook):
         raise
 
     # Submit log to Log Detective and await the results.
-    response = await submit_log_to_llm(preprocessed_log)
+    log_text = preprocessed_log.read().decode(encoding="utf-8")
+    staged_response = await perform_staged_analysis(log_text=log_text)
     preprocessed_log.close()
 
     # Add the Log Detective response as a comment to the merge request
-    await comment_on_mr(merge_request_id, response)
+    await comment_on_mr(merge_request_id, staged_response)
 
 
 class LogsTooLargeError(RuntimeError):
@@ -584,15 +585,11 @@ async def check_artifacts_file_size(job):
     return content_length <= SERVER_CONFIG.gitlab.max_artifact_size
 
 
-async def submit_log_to_llm(log):
-    """Stream the log to the LLM for processing"""
-    # TODO: query the LLM with the log contents  # pylint: disable=fixme
-    # This function will be implemented later; right now it does nothing.
-    LOG.debug("Log contents:\n%s", log.read())
-    return ""
-
-
-async def comment_on_mr(merge_request_id: int, response: str):  # pylint: disable=unused-argument
+async def comment_on_mr(merge_request_id: int, response: StagedResponse):
     """Add the Log Detective response as a comment to the merge request"""
-    # TODO: Implement this  # pylint: disable=fixme
-    pass  # pylint: disable=unnecessary-pass
+    LOG.debug("Primary Explanation for MR %d: %s", merge_request_id, response.explanation.text)
+
+    for snippet in response.snippets:
+        LOG.debug("")
+        LOG.debug("%d: %s", snippet.line_number, snippet.text)
+        LOG.debug("%s", snippet.explanation)
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -1,9 +1,12 @@
 import asyncio
+import gitlab.v4
+import gitlab.v4.objects
+import jinja2
 import json
 import os
 import re
 import zipfile
-from pathlib import PurePath
+from pathlib import Path, PurePath
 from tempfile import TemporaryFile
 from typing import List, Annotated, Tuple, Dict, Any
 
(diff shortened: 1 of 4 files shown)


sgallagher:
https://gitlab.com/redhat/centos-stream/rpms/osci-internal-test-package/-/merge_requests/820#note_2427683013 was produced by my development Log Detective server.

logdetective/server/server.py
```diff
@@ -600,20 +599,30 @@ async def check_artifacts_file_size(job):
...
     # Get the formatted comment.
     comment = await generate_mr_comment(job, log_url, response)
 
+    # Look up the merge request
+    merge_request = await asyncio.to_thread(
+        project.mergerequests.get, merge_request_iid
+    )
+
     # Submit a new comment to the Merge Request using the Gitlab API
+    await asyncio.to_thread(merge_request.discussions.create, {"body": comment})
```
majamassarini, line 625: I imagine that the `python-gitlab` objects don't support asynchronous methods, which is unfortunate. In my experience, working with threads in an asynchronous application doesn't scale well. The `ThreadPoolExecutor` typically has only a limited number of workers, which can become a bottleneck.

logdetective/server/server.py
```diff
@@ -588,15 +589,11 @@ async def check_artifacts_file_size(job):
...
-    LOG.debug("Log contents:\n%s", log.read())
-    return ""
-
-
-async def comment_on_mr(merge_request_id: int, response: str):  # pylint: disable=unused-argument
+async def comment_on_mr(merge_request_id: int, response: StagedResponse):
     """Add the Log Detective response as a comment to the merge request"""
-    # TODO: Implement this  # pylint: disable=fixme
-    pass  # pylint: disable=unnecessary-pass
+    LOG.debug("Primary Explanation for MR %d: %s", merge_request_id, response.explanation.text)
```
TomasTomecek: This will be very helpful for MVP to see which MR we are working with and what exactly the explanation is.

At the same time I hope we'll be able to come up with a longterm solution for logs which will be easy to navigate - our current log will be a giant wall of text. Even now when we are logging the raw `/completions` calls with probabilities.

logdetective/server/server.py
```diff
@@ -588,15 +589,11 @@ async def check_artifacts_file_size(job):
...
+
+    for snippet in response.snippets:
+        LOG.debug("")
+        LOG.debug("%d: %s", snippet.line_number, snippet.text)
+        LOG.debug("%s", snippet.explanation)
```
TomasTomecek: Very nice formatting!

logdetective/server/server.py
```diff
@@ -558,11 +566,13 @@ async def retrieve_and_preprocess_koji_logs(job):
 
     LOG.debug("Failed architecture: %s", failed_arch)
 
-    log_path = failed_arches[failed_arch]
-    LOG.debug("Returning contents of %s", log_path)
+    log_path = failed_arches[failed_arch].as_posix()
+
+    log_url = f"{SERVER_CONFIG.gitlab.api_url}/projects/{job.project_id}/jobs/{job.id}/artifacts/{log_path}"
```
TomasTomecek: nit: I'm surprised the python gitlab bindings don't provide a function to construct this URL

logdetective/server/templates/gitlab_comment.md.j2
```diff
@@ -0,0 +1,66 @@
+The package {{ package }} failed to build, here is a possible explanation why.
+
+Please know that the explanation was provided by AI and may be incorrect.
+In this case, we are {{ certainty }}% certain of the response :neutral_face:.
```
TomasTomecek, line 4: I'm assuming we'll also template/configure the emoji, right? :grin:

TomasTomecek (approved):
I couldn't find any issues in the code, this is a surgically precise pull request :)

logdetective/server/server.py
sgallagher: Yeah, it's unfortunate, but I think we'll be alright in this case, as each of these tasks are relatively short and self-contained. If we were trying to run long-term tasks in the threads, I'd definitely find a new approach.

The obvious alternative would be for us to essentially reimplement the python-gitlab library and build it atop the `aiohttp` or `httpx` module, as is proposed in https://github.com/python-gitlab/python-gitlab/issues/1025 but I don't think we realistically have the time to do that right now.

We could also look at switching over to the GraphQL API for Gitlab and using `aiographql`, but that is a significant change in the API behavior.

I'll open a ticket to investigate whether either of those strategies would be worth the time to implement.

logdetective/server/server.py
sgallagher: I have another half-formed merge request in the works, but my suggestion is that we should adjust to using `INFO` for things we actually want to have visible in the stdout/stderr log and then just shunt `DEBUG` to the `logdetective.log` file. That's what I do in bureaucrabot/check-tickets: https://gitlab.com/redhat/centos-stream/ci-cd/dist-git-gating-tests/-/blob/latest/bureaucrabot/__init__.py?ref_type=heads#L190

logdetective/server/server.py
sgallagher: They allow us to directly retrieve the contents of the file, so providing us the URL itself is probably not something the python-gitlab developers thought would be particularly useful.

Also, that's not a direct URL; I think that returns a 302 redirect to a load-balanced, geolocated file server.

logdetective/server/templates/gitlab_comment.md.j2
sgallagher: Honestly, I hadn't even considered that. @FrostyX didn't templatize that, so I didn't change it.

(And frankly, given the 94% or more confidence it's offering for some _very_ poor answers, the :neutral_face: might be the right choice!)

logdetective/server/server.py
majamassarini: I think that creating a ticket for investigate this **in future** is perfect! I agree that right now this is not a problem at all :) I was just a bit worried for the future. Thanks.

logdetective/server/server.py
TomasTomecek: that's a solid suggestion
I'd love to use splunk (or something of that magnitude) in the future
your proposal will work well short/mid term IMO

logdetective/server/templates/gitlab_comment.md.j2
sgallagher: Too late, I just pushed another patch to this set that implements it. Thanks for the pointer, @FrostyX

PR #165, opened 2025-04-01
Use HTTPS in production

FrostyX:
Fix LD-35


TomasTomecek (approved):
Wow, very nice!! I did not expect that one command would resolve this.

jpodivin (approved):

PR #167, opened 2025-04-01
Loosen the merge_request_id regex

sgallagher:
It turns out that sometimes the format is `refs/merge-requests/<ID>/merge` and sometimes it is `refs/merge-requests/<ID>/head`. There may be others, so we'll loosen the check a bit.


FrostyX (approved):

TomasTomecek:
Oh my, so much for consistency... Well done figuring it out!

sgallagher:
> Oh my, so much for consistency... Well done figuring it out!

I think this is related to another bug we see where sometimes the Merge Request pipeline has a merge commit at the HEAD and sometimes it doesn't.

PR #177, opened 2025-04-02
Plot a graph for analyze (endpoint) requests over time

majamassarini:
Shows how many requests, for the analyze endpoint, we had in the specified period of time.

Examples queries:

http GET "localhost:8080/metrics/analyze/requests" > /tmp/plot.svg  # defaults to the last two days
http GET "localhost:8080/metrics/analyze/requests?hours=5" > /tmp/plot_hours.svg
http GET "localhost:8080/metrics/analyze/requests?days=5" > /tmp/plot_days.svg
http GET "localhost:8080/metrics/analyze/requests?weeks=5" > /tmp/plot_weeks.svg

Examples of plots generated using the unit tests:

tmpuifpmdjf
tmpcuhouf67
tmp488wkb00

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,17 +294,34 @@ HTTPS certificate generated through:
 certbot certonly --standalone -d logdetective01.fedorainfracloud.org
 ```
 
-Query metrics
--------------
+Querying statistics
+-------------------
 
-You can query how many requests there were in the server
-in the last given period of time using the following command:
+You can retrieve statistics about server requests over a specified time period
+using either the `curl` command or the `http` command (provided by the `httpie` package).
 
-http GET "localhost:8080/metrics/analyze/requests" > /tmp/plot.svg  # defaults to the last two days
+When no time period is specified, the query defaults to the last 2 days:
+
+```
+http GET "localhost:8080/metrics/analyze/requests" > /tmp/plot.svg
+curl "localhost:8080/metrics/analyze/staged/requests" > /tmp/plot.svg
+```
+
+You can specify the time period in hours, days, or weeks.
+The time period:
+
+ - cannot be less than one hour
+ - cannot be negative
+ - ends at the current time (when the query is made)
+ - starts at the specified time interval before the current time.
+
+Examples:
+
+```
 http GET "localhost:8080/metrics/analyze/requests?hours=5" > /tmp/plot_hours.svg
 http GET "localhost:8080/metrics/analyze/requests?days=5" > /tmp/plot_days.svg
 http GET "localhost:8080/metrics/analyze/requests?weeks=5" > /tmp/plot_weeks.svg
-
+```
 
 License
 -------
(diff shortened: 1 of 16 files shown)


logdetective/server/plot.py
```diff
@@ -0,0 +1,205 @@
...
+        f"Requests received ({start_time.strftime(plot_def.time_format)} "
+        f"to {end_time.strftime(plot_def.time_format)})"
+    )
+
+    lines1, labels1 = ax1.get_legend_handles_labels()
+    lines2, labels2 = ax2.get_legend_handles_labels()
+    ax1.legend(lines1 + lines2, labels1 + labels2, loc="center")
+
+    matplotlib.pyplot.tight_layout()
+    matplotlib.pyplot.show()
```
jpodivin, line 203: Do we really need to call show here?

majamassarini: no 🙈, good catch!

logdetective/server/plot.py
```diff
@@ -0,0 +1,206 @@
...
+        f"to {end_time.strftime(plot_def.time_format)})"
+    )
+
+    lines1, labels1 = ax1.get_legend_handles_labels()
+    lines2, labels2 = ax2.get_legend_handles_labels()
+    ax1.legend(lines1 + lines2, labels1 + labels2, loc="center")
+
+    matplotlib.pyplot.tight_layout()
+
+    return fig
```
TomasTomecek, line 206: Maja, this is super impressive!! I expected a "stupid" local script that we would call in some funky way. Having this integrated directly in the site is mindblowing!

README.md
```diff
@@ -294,6 +294,17 @@ HTTPS certificate generated through:
 certbot certonly --standalone -d logdetective01.fedorainfracloud.org
 ```
 
+Query metrics
+-------------
+
+You can query how many requests there were in the server
+in the last given period of time using the following command:
+
+http GET "localhost:8080/metrics/analyze/requests" > /tmp/plot.svg  # defaults to the last two days
```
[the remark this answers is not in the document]
majamassarini: sure, I also added an example with the curl command, I forgot that we were using it for the examples 😅

logdetective/server/plot.py
majamassarini: Thanks Tomas, I think this approach is quicker and more convenient but I am not sure this is what we want. @jpodivin I am mistaken or you are concerned about exposing the metrics in the server? (I got this feeling in the yesterday standup). I can move this same code in a script if you prefer.

jpodivin: I'm somewhat concerned about people spamming it, thus causing issues with response times for the server as a whole. That being said, I understand that it may be a premature worry. I suppose we can merge this and later maybe split this off into it's own server or add authentication.

jpodivin (approved):

logdetective/server/plot.py
majamassarini: For the future I believe we should send the metrics to Grafana (using the Pushgateway) or something similar. I believe these plots and their endpoints can be trow away when we will have the necessary infrastructure for using Grafana.

PR #182, opened 2025-04-03
Copy run_server.sh and gunicorn configs to the image

FrostyX:
Otherwise, the image refuses to start outside our docker-compose setup that bindmounts the files in.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -31,5 +31,7 @@ RUN pip3 install llama_cpp_python==0.2.85 sse-starlette starlette-context \
 COPY ./logdetective/ /src/logdetective/
 COPY ./alembic.ini /src/alembic.ini
 COPY ./alembic /src/alembic
+COPY ./files /src/files
+COPY ./server /src/server
 
 WORKDIR /src


Containerfile
```diff
@@ -31,5 +31,7 @@ RUN pip3 install llama_cpp_python==0.2.85 sse-starlette starlette-context \
 COPY ./logdetective/ /src/logdetective/
 COPY ./alembic.ini /src/alembic.ini
 COPY ./alembic /src/alembic
+COPY ./files /src/files
+COPY ./server /src/server
```
sgallagher, line 35: I think we still need to bind-mount the `server/config.yml` (so an OpenShift Secret can provide it) since it will contain API keys for both Gitlab API and the LLM.

sgallagher (changes requested):

Containerfile
FrostyX: That's a good point and I agree with you. So if you want, I can specifically copy only the `gunicorn-*.config.py` files from the `server` directory.

My plan was to copy everything and then create a `config.yml` as an openshift secret and specify `LOGDETECTIVE_SERVER_CONF` to use the secret config, not the default one.

FrostyX:
I am going to merge, so that I can move forward with the OpenShift deployment. We all agree that the actual production config needs to be an OpenShift secret.

PR #183, opened 2025-04-04
Adding settings enabling OpenScanHub checks for our package

jpodivin:
I've added suggested baseline settings allowing OpenScanHub to run on our code.
This would add one more layer to our code quality assurance automation. Provided that it works as intended.

If not, or if it produces too many false positives, we should disable it and return to previous setup.
And report the issues of course.

### DIFF
--- a/packit.yaml
+++ b/packit.yaml
@@ -26,6 +26,11 @@ jobs:
     trigger: pull_request
     targets:
       - fedora-all
+  - job: copr_build
+    trigger: commit
+    branch: main
+    targets:
+      - fedora-all
   - job: tests
     trigger: pull_request
     targets:
--- a/packit.yaml
+++ b/packit.yaml
@@ -26,11 +26,13 @@ jobs:
     trigger: pull_request
     targets:
       - fedora-all
+  # Run build on commit with OpenScanHub checks
   - job: copr_build
     trigger: commit
     branch: main
     targets:
       - fedora-all
+    osh_diff_scan_after_copr_build: true
   - job: tests
     trigger: pull_request
     targets:


sgallagher:
I assume we're relying on the default behavior described at https://packit.dev/docs/configuration#osh_diff_scan_after_copr_build

Could you either make that config option explicit or add a comment explaining it (or both), please? There's no indication what this change is doing, otherwise.

siteshwar:
You may also want to enable `pylint` through [csmock_args](https://packit.dev/docs/configuration#csmock_args). It is not enabled by default.

jpodivin:
> You may also want to enable `pylint` through [csmock_args](https://packit.dev/docs/configuration#csmock_args). It is not enabled by default.

I've understood that it's kind of flaky, producing too many false positives, but I guess we can always revert.

kdudka:
@TomasTomecek Looking at the diff, the check is triggered by a commit to `main` branch but not on a pull request.

siteshwar:
> I'm confused, how come there are no links to OSH neither here in github checks neither in Packit dashboard.

This pull request should be merged and the target branch should have a successful copr build for the first differential scan to be submitted via Packit.

siteshwar:
> Also, according to csmock docks, the `--pylint-scan-install` should be enabled by default. https://www.mankier.com/1/csmock#--pylint-scan-install

It is disabled on the Fedora instance. Please use `-t="pylint"` as `csmock args` to enable it. Thanks!

kdudka:
@siteshwar Good point.  `--pylint-scan-install` is an option of the `pylint` plug-in.  Although the flag is enabled by default, the `pylint` plug-in itself needs to be explicitly enabled first.

siteshwar:
> It is disabled on the Fedora instance. Please use `-t="pylint"` as `csmock args` to enable it. Thanks!

On a second thought, you should also enable `shellcheck` as you have a [shellscript](https://github.com/fedora-copr/logdetective/blob/29b2315f4a87d09a4931adc7e52793ab6f617948/files/run_server.sh) in this repository.  I triggered a scan manually and you can see the [full scan results](https://github.com/fedora-copr/logdetective/blob/29b2315f4a87d09a4931adc7e52793ab6f617948/files/run_server.sh), although the warnings seem harmless.

sgallagher (approved):

PR #184, opened 2025-04-04
Prompts from files

jpodivin:
Instead of having only hard coded prompts, we will allow users to supply their own in a file.
The constants will still serve as sensible defaults. Used when no file is provided.

This PR introduces a file for prompts which is separate from server config, as it will be used by both CLI and server implementations of Log Detective.


logdetective/prompts.yml
```diff
@@ -0,0 +1,50 @@
+prompt_template: |
```
TomasTomecek, line 1: since we already use jinja2, how about using it for templating here as well?

jpodivin: It's a template, but not a jinja template. We are using `format` function for substitution. Using jinja template here would mean splitting `prompts.yml` into multiple files, then passing them to a rendering engine, and finally using resultant strings.

When we only need to deal with one variable, that seems like an overkill.

TomasTomecek: okay, that's fair; we can always adapt jinja later if we wanted to

logdetective/utils.py
```diff
@@ -199,3 +202,15 @@ def validate_url(url: str) -> bool:
     if not (result.path or result.netloc):
         return False
     return True
+
+
+def load_prompts(path: str | None) -> PromptConfig:
```
TomasTomecek, line 207: I'd like to ask you to create at least one test case for this function.

TomasTomecek (changes requested):
Looks solid to me! Please create a test case for the newly introduced function.

logdetective/utils.py
jpodivin: Thanks for reminding me. Tests incoming.

TomasTomecek (approved):

TomasTomecek:
@jpodivin feel free to merge when you feel it's ready to be deployed

PR #186, opened 2025-04-05
Make HTTPS certificates directory configurable

FrostyX:
This will make our OpenShift deployment easier.


FrostyX:
I am going to merge without a review because it blocks my OpenShift experiments. We can change this later.

Issue #93, opened 2024-12-03
Refactor prompt construction code


jpodivin:
Prompts are constructed out of built-in strings, often used as f-string. I would prefer them to be loaded from a file.
Ideally yaml, or other similar, easily parsable format.

TomasTomecek:
Agreed! Changing prompts without editing code sounds like a great idea - that will make experimenting much easier.

abadger:
This appears to be implemented in https://github.com/fedora-copr/logdetective/pull/184 and can be closed.

PR #188, opened 2025-04-05
Refactor prompts into YAML

san-118:
This PR addresses issue #93 by moving all prompt templates to a YAML configuration file (prompts.yaml). Prompts are now loaded using constants.py by cached YAML parsing. This makes prompts easier to maintain, edit, and experiment without modifying code directly.

- Prompts now live in prompts.yaml .
- All prompt usage updated to import from constants.py
- Clean separation of config and logic
- No changes needed in utils.py or other processing code

Tested locally and existing functionality works as expected.

### DIFF
--- a/logdetective/prompts.yaml
+++ b/logdetective/prompts.yaml
@@ -1,0 +1,58 @@
+default_advisor: fedora-copr/Mistral-7B-Instruct-v0.2-GGUF
+
+prompts:
+  analyze_snippets:
+    template: |
+      Given following log snippets, and nothing else, explain what failure, if any, occured during build of this package.
+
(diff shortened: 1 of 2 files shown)


TomasTomecek:
@san-118 thank you for your contribution!

We already have a PR open from @jpodivin in #184 that addresses the issue and goes further than this PR. 

We are in a rapid development mode in Log Detective so I strongly suggest to check with us in an issue if you wanna pick one up so we don't waste your time.

I'm sorry but we'll have to close this PR.

logdetective/prompts.yaml
```diff
@@ -0,0 +1,58 @@
+default_advisor: fedora-copr/Mistral-7B-Instruct-v0.2-GGUF
```
TomasTomecek, line 1: it's odd this configuration value made it to this file

PR #202, opened 2025-04-09
Put email notifications on a diet

sgallagher:
## Make Gitlab comment snippet display cleaner

Fixes the extra newline at the start of each bullet.

Also pulls the line number out of the `<code>` block and emboldens
it for visibility.

## Separate comment template into short and full

In order to not flood users with long emails, we will send only the
summary out in the initial comment and then submit an update after
five seconds. This works because Gitlab doesn't send an email to
notify about edits to comments, only the initial comment creation.

It includes a five-second sleep in order to ensure that the
creation and edit don't get bundled together closely enough that
they go out as a single email.


TomasTomecek (approved):
Very nice!! Feel free to merge :)

PR #203, opened 2025-04-09
Allow setting different log levels for STDOUT and logfile

FrostyX:
Fix LD-42

This is not the most universal solution but I doubt we will use more handlers in the future, so I'd prefer keeping this simple.


FrostyX:
@sgallagher you were mostly interested in this. What do you think?

sgallagher (approved):

TomasTomecek (approved):

PR #204, opened 2025-04-09
Specify model and max_tokens when querying models

FrostyX:
When using docker-compose, our inference container gets these from our `.env` file (see `LLAMA_ARG_MODEL`). However, in production we use models.corp, so these are not being set anymore. We need to specify them in our request.

### DIFF
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -95,6 +95,7 @@ class InferenceConfig(BaseModel):
     )
     url: str = ""
     api_token: str = ""
+    model: str = ""
 
     def __init__(self, data: Optional[dict] = None):
         super().__init__()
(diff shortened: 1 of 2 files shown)


FrostyX:
@TomasTomecek can you please take a look?
When I configured all of these locally, set the `inference.api_token` and pointed the `inference.url` to the models.corp URL, everything worked. So this should work in production too.

jpodivin (approved):
Yes, this should work. We may have to adjust the `max_tokens` values. Depending on the API. `-1` should be fine, but OpenAI API is pretty vague about that.

FrostyX:
>  We may have to adjust the max_tokens values. Depending on the API. -1 should be fine, but OpenAI API is pretty vague about that.

I required a value higher than `1`:

```
{"detail":"Something went wrong while getting a response from the llama server: [400] {\"object\":\"error\",\"message\":\"max_tokens must be at least 1, got -1.\",\"type\":\"BadRequestError\",\"param\":null,\"code\":400}"}
```

But once I set the value in `config.yml` it worked fine.

TomasTomecek (approved):
Very nice!! I also struggled with the max_tokens value while experimenting.

PR #210, opened 2025-04-10
Do not block FastAPI server when submitting to llm

majamassarini:
requests.post is blocking, handle it in a separate thread.

### DIFF
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -1,8 +1,11 @@
 import asyncio
+import functools
 import json
 import os
 import re
 import zipfile
+from concurrent.futures import ThreadPoolExecutor
+from contextlib import asynccontextmanager
 from pathlib import Path, PurePath
 from tempfile import TemporaryFile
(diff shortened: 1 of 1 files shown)


sgallagher (commented):
Why go this route instead of `asyncio.to_thread()` like we do in other places?

sgallagher (commented):
Alternatively, aiohttp might be a better option here, as it's natively asyncio-aware.

majamassarini:
> Alternatively, aiohttp might be a better option here, as it's natively asyncio-aware.

I completely agree. I said this ways a quick and dirty solution in [chat](https://redhat-internal.slack.com/archives/C0671USMZJN/p1744022215830869) and I proposed to move to another http async library for a better solution. I am just wondering if now is the right time for moving?

TomasTomecek:
Since we are now 100% on aiohttp, let's close this; thanks Maja for sparking the effort to go fully async :)

PR #211, opened 2025-04-10
more examples of prompts

xsuchy:
This can give users ideas how to tune it to their needs.

### DIFF
--- a/logdetective/prompts-summary-only.yml
+++ b/logdetective/prompts-summary-only.yml
@@ -1,0 +1,13 @@
+# This file is intended for customization of prompts
+# It is used only in server mode.
+# On command line you have to load it using --prompts
+# The defaults are stored in constants.py
+
+prompt_template: |
+  Given following log snippets, and nothing else, explain what failure, if any, occured during build of this package.
+
+  Provide concise, one paragraph long, summary describing the problem of most probable culprit and recommend solution to fix it.
+
+  Snippets:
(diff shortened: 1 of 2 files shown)


logdetective/prompts-summary-only.yml
```diff
@@ -0,0 +1,9 @@
+# This file is intended for customization of prompts
```
jpodivin, line 1: This one won't work. You need the `{}` for replacement field.

jpodivin (changes requested):

logdetective/prompts-summary-only.yml
xsuchy: ??? I do not want to replace anything.

jpodivin: The problem is how the prompt is meant to be used.
This prompt eventually reaches https://github.com/fedora-copr/logdetective/blob/main/logdetective/utils.py#L129 where it is used with `format` method.

If there is no `{}` replacement field. [1] There is no place for insertion of formatted log.

I realize this should have been specified from the start. But now it's at least in readme.[2]

[1] https://docs.python.org/3/library/string.html#format-string-syntax 
[2] https://github.com/fedora-copr/logdetective/tree/main?tab=readme-ov-file#system-prompts

xsuchy:
Updated. Please review.

PR #214, opened 2025-04-10
Draft commit for db retries

majamassarini:


xsuchy:
What about https://pypi.org/project/backoff/ ?

majamassarini:
> What about https://pypi.org/project/backoff/ ?

Right, even better, I forgot about it 😅 Tomorrow I will substitute it in this PR. Now I am still trying to make alembic work, for creating the new db schema, after the latest changes in our composes files.

PR #215, opened 2025-04-10
replace requests with aiohttp

TomasTomecek:
Commit msg:

    replace requests with aiohttp

    requests is no longer being imported in our codebase

    we now do all http requests with aiohttp

    this commit also contains a test case for HTTP communication with
    /completions endpoints by starting a local llama-server; sadly it's not
    available via PyPI so we need to figure out how to lunch the server in
    CI


sgallagher (changes requested):
A few points:

1. It's not a great idea to create a session as a context manager every time you make a request; it's unnecessary overhead that could be done a single time at the start of the server. You can also set certain defaults (like the timeout, see next point) on the session object to avoid needing to include them in the `get()` and `post()` calls.
2. The `timeout` field can (and should) actually take a `aiohttp.ClientTimeout` object which can have different connect and total timeouts. We should probably retain the current timeouts.

TomasTomecek:
I've incorporated the same lifespan pattern here. It was challgening to run tests with it because the FastAPI app was not initialized. I welcome suggestions how to write this better.

majamassarini:
(same message as in chat :))
I have pushed two commits into your PR, feel free to remove both of them if you want. 
In the first one I have used another dependency, just in tests, to test the asyncio method without running the server. In the second I have substituted the app: FastAPI reference with the http: aiohttp.ClientSession object (mainly because otherwise pylint complains a lot, but also because I think it is better not to bring about the app object itself). 
For both **feel free to remove the commits**, they are just a suggestion :slightly_smiling_face: I haven't tested the PR locally yet.

TomasTomecek:
@majamassarini I incorporated your changes and ported all new tests to that paradigm, please have a look

majamassarini:
@TomasTomecek I tested it locally, I realized I had done some mistakes in my previous commit (I have added a new commit with the fix, I wasn't able to push a rewrite of the git history 😅). The only endpoint I tested is working for me.

TomasTomecek:
@majamassarini I merged those commits together :) thanks for noticing

I also realized I had to change the CLI code to actually invoke the async main properly, I verified locally it works

logdetective/server/server.py
```diff
@@ -147,40 +160,40 @@ async def submit_to_llm_endpoint(
...
     """
+    LOG.debug("async request %s headers=%s data=%s", url, headers, data)
     try:
-        # Expects llama-cpp server to run on LLM_CPP_SERVER_ADDRESS:LLM_CPP_SERVER_PORT
-        response = requests.post(
+        response = await http.post(
             url,
             headers=headers,
-            data=json.dumps(data),
+            # llama-server doesn't accept the encoding that data param does
```
sgallagher, line 168: This comment is confusing. What is the encoding that the data parameter uses, why doesn't llama-server handle it, and what are we doing to work around it?

sgallagher (approved):
Other than a minor complaint about a comment, I think this is looking great!

logdetective/server/server.py
TomasTomecek: I actually don't know what exactly has happened but the encoding that aiohttp performed using `data=` always failed with the server not being able to json-decode the request (unexpected char at position 1), even when the header was explicitly set to json; no idea what problem these two rascals had

once I changed it to `json=data`, all worked smoothly

now that I read this, let me update the comment

sgallagher (approved):

Issue #213, opened 2025-04-10
db connection closed unexpectedly


xsuchy:
I am trying to do some performance test and when I do:

```
$ cat /tmp/data.json
{"url":"https://kojipkgs.fedoraproject.org//work/tasks/4967/130854967/build.log"}
$ ab  -n 10 -c 4 -T application/json -P /tmp/data.json https://logdetective01.fedorainfracloud.org/analyze/staged
```

It fails very fast. And the log has traceback that `db connection closed unexpectedly`. This seems to be related to `analyze` decorators added recently. (cc @majamassarini )

BTW I had 100% reproducibility with the steps above.

FrostyX:
Traceback from the logdetective01 server:

```
6561c68ab1f7 2025-04-10 09:47:49,763 INFO sqlalchemy.engine.Engine BEGIN (implicit)
6561c68ab1f7 2025-04-10 09:47:49,763 INFO sqlalchemy.engine.Engine INSERT INTO analyze_request_metrics (endpoint, request_received_at, log_url, response_sent_at, response_length, response_certainty) VALUES (%(endpoint)s, %
(request_received_at)s, %(log_url)s, %(response_sent_at)s, %(response_length)s, %(response_certainty)s) RETURNING analyze_request_metrics.id
6561c68ab1f7 2025-04-10 09:47:49,763 INFO sqlalchemy.engine.Engine [cached since 7.698e+04s ago] {'endpoint': 'ANALYZE_STAGED', 'request_received_at': datetime.datetime(2025, 4, 10, 9, 47, 49, 763492, tzinfo=datetime.timez
one.utc), 'log_url': 'https://kojipkgs.fedoraproject.org//work/tasks/4967/130854967/build.log', 'response_sent_at': None, 'response_length': None, 'response_certainty': None}
6561c68ab1f7 2025-04-10 09:47:49,764 - WARNING - Exception while working with database: (psycopg2.OperationalError) server closed the connection unexpectedly
6561c68ab1f7    This probably means the server terminated abnormally
6561c68ab1f7    before or while processing the request.
6561c68ab1f7
6561c68ab1f7 [SQL: INSERT INTO analyze_request_metrics (endpoint, request_received_at, log_url, response_sent_at, response_length, response_certainty) VALUES (%(endpoint)s, %(request_received_at)s, %(log_url)s, %(response_
sent_at)s, %(response_length)s, %(response_certainty)s) RETURNING analyze_request_metrics.id]
[...]

PR #217, opened 2025-04-11
Add merge_requests, comments and reactions tables

majamassarini:


logdetective/server/database/models/__init__.py
```diff
@@ -0,0 +1,12 @@
+from logdetective.server.database.base import Base
+from logdetective.server.database.models.merge_requests import MergeRequests, Comments, Reactions
+from logdetective.server.database.models.metrics import AnalyzeRequestMetrics, EndpointType
+
+__all__ = [
```
majamassarini, line 5: @jpodivin I know you don't like these imports but otherwise we should import every new database model class explicitly [here](https://github.com/fedora-copr/logdetective/blob/0709391db1ee9bda84ecfcd6d4cd64861af8dd43/alembic/env.py#L8) every time we create a new one (to use the alembic autogenerate feature). 
Or we should go back as it was before, having only one `models.py` file, but it would be already more than 1000 lines long.

.gitignore
```diff
@@ -157,4 +157,4 @@ dmypy.json
 cython_debug/
 
 # Model files
-models/
+/models/
```
TomasTomecek, line 160: wow, this is a great catch!

alembic/versions/6e1fe3f1eccf_add_merge_requests_comments_reactions_.py
```diff
@@ -0,0 +1,118 @@
...
+    op.create_table(
+        "merge_requests",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column(
+            "mr_id", sa.Integer(), nullable=False, comment="The merge request id"
+        ),
+        sa.Column("project_id", sa.Integer(), nullable=False, comment="The project id"),
+        sa.Column("job_id", sa.Integer(), nullable=False, comment="The job id"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("mr_id", "project_id", "job_id", name="uix_mr_project_job"),
```
TomasTomecek, line 34: just to be clear: when we "rerun" the log detective analysis, will it get a new job id?

alembic/versions/6e1fe3f1eccf_add_merge_requests_comments_reactions_.py
```diff
@@ -0,0 +1,118 @@
...
+    op.create_table(
+        "reactions",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column(
+            "comment_id",
+            sa.Integer(),
+            nullable=False,
+            comment="The associated comment id",
+        ),
+        sa.Column(
```
TomasTomecek, line 88: I'm just wondering here; so if a comment has 5 :+1:, does that mean it will have 5 distinct records? if so, we should add something here to distinguish between them, i.e. datetime when the reaction was recorded

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,389 @@
...
+    created by logdetective in merge requests."""
+
+    __tablename__ = "comments"
+
+    id = Column(Integer, primary_key=True)
+    merge_request_id = Column(
+        Integer,
+        ForeignKey("merge_requests.id"),
+        nullable=False,
+        unique=True,  # 1 comment for 1 job
```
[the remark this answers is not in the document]
TomasTomecek: oh, I made the same comment on the migration above :see_no_evil:

.gitignore
jpodivin: Previously we didn't have a `models` dir on lower levels so it could be avoided.

pyproject.toml
```diff
@@ -49,10 +49,11 @@ psycopg2-binary = {version = "^2.9.9", optional = true }
 psycopg2 = {version = "^2.9.9", optional = true }
 alembic = {version = "^1.13.3", optional = true }
 matplotlib = {version = "^3.8.4", optional = true }
+backoff = {version = "2.2.1", optional = true }
```
jpodivin, line 52: This is may become a problem for us. The library is no longer maintained.[1] While it should have everything we need at this point, we need to be aware of the fact.

[1] https://github.com/litl/backoff/issues/211

logdetective/server/database/models/metrics.py
```diff
@@ -75,20 +84,20 @@ def create(
...
+    @backoff.on_exception(backoff.expo, OperationalError, max_tries=DB_MAX_RETRIES)
     def update(
         cls,
         id_: int,
         response_sent_at: datetime,
         response_length: int,
         response_certainty: float,
     ) -> None:
-        """Update an AnalyzeRequestMetrics line
+        """Update an  line
```
jpodivin, line 100: nit `s/an  line/a line/`

logdetective/server/database/models/metrics.py
```diff
@@ -97,6 +106,47 @@ def update(
...
+        with transaction(commit=True) as session:
+            session.merge(self)
+
+    @classmethod
+    def get_requests_metrics_for_mr(
+        cls,
+        mr_id: int,
+        project_id: int,
+        job_id: int,
+    ) -> Optional["AnalyzeRequestMetrics"]:
```
jpodivin, line 128: I'm not sure what this means. Does it return `AnalyzeRequestMetrics` only sometimes, what is returned in other cases if not? Line 146 would indicate that result should be a list of some kind.

majamassarini: good catch, it should be a list.

pyproject.toml
majamassarini: we discussed at the standup that pinning down the version should prevent us from future problems.

logdetective/server/database/models/__init__.py
majamassarini: I don't like to use alias if not strictly needed. I would keep the code as it is if you don't prefer to have just a single `models.py` file.

alembic/env.py
```diff
@@ -23,6 +23,8 @@
 # target_metadata = mymodel.Base.metadata
 target_metadata = models.Base.metadata
 
+print("Tables detected by Alembic:", [table for table in target_metadata.tables.keys()])
```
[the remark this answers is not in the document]
majamassarini: Sometime `alembic autogenerate` doesn't create the migrations properly. This happens because there are no imports for the new models, with this print I can see if the model has been properly imported. I agree it is not needed, I will leave it here (for when someone needs to debug the problem) but I will **comment it** out.

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,396 @@
...
+    ForeignKey,
+    UniqueConstraint,
+    desc,
+)
+from sqlalchemy.orm import relationship
+from sqlalchemy.exc import OperationalError
+from logdetective.server.database.base import Base, transaction, DB_MAX_RETRIES
+
+
+class MergeRequests(Base):
```
sgallagher, line 20: We should probably call this `GitlabMergeRequests` since the relationships here are GItlab-specific. There will likely be significant differences in the way that other forges handle things that we may need to account for in the future.

sgallagher, line 20: It's possible over an extended lifetime of Log Detective that we will need to handle multiple Gitlab instances hosted in different places. We should probably include an instance ID here (and create another table that provides that instance's base URL, such as `https://gitlab.com` or `https://gitlab.example.com`.

This will also mean changing the uniqueness constraints to include this new field.

sgallagher (changes requested):

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,396 @@
...
+        index=True,
+        comment="The merge request id",
+    )
+    project_id = Column(
+        Integer,
+        nullable=False,
+        index=True,
+        comment="The project id",
+    )
+    job_id = Column(
```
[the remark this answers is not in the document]
majamassarini: And probably the same applies also to the project_id?

majamassarini:
@sgallagher I should have applied all your suggestions. The only thing a bit different is that I have created an Enum for the Forges and their urls, instead of a table, I thought that we will add them manually and not dynamically.

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,462 @@
...
+from sqlalchemy.orm import relationship
+from logdetective.server.database.base import Base, transaction
+
+
+class Forge(str, enum.Enum):
+    """List of forges managed by logdetective"""
+
+    gitlab_com = "https://gitlab.com"  # pylint: disable=(invalid-name)
+    gitlab_cee_redhat_com = "http://gitlab.cee.redhat.com/"  # pylint: disable=(invalid-name)
+
```
sgallagher, line 27: On the one hand, I don't like having this as an `Enum`, because it inherently means that we need to modify the database schema in order to add another possible entry. An `Enum` is really intended to constrain values when there's a known, finite set that is valid. Whereas the set of possible Gitlab instances is potentially infinite, even if we only expect to handle a few.

On the other hand, it _does_ provide some protection from typographical errors (or someone trying to fill our DB table by sending us random `X-Gitlab-Server` headers to waste storage).

I suppose we can go with an Enum for now and fix things later if we grow to the point where it's a problem.

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,462 @@
...
+
+class GitlabMergeRequests(Base):
```
sgallagher, line 29: We might want to call this `GitlabMergeRequestJobs` instead, since the `(forge, project, mr)` combination may appear multiple times, representing multiple job runs of the same merge request.

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+class Comments(Base):
+    """Store details for comments
+    created by logdetective in merge requests."""
+
+    __tablename__ = "comments"
+
+    id = Column(Integer, primary_key=True)
+    merge_request_id = Column(
```
sgallagher, line 165: Same as above; we're really associating with a particular job on a merge request, not the merge request itself.

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,462 @@
...
+        Integer,
+        ForeignKey("gitlab_merge_requests.id"),
+        nullable=False,
+        unique=True,  # 1 comment for 1 job
+        index=True,
+        comment="The associated merge request id",
+    )
+    comment_id = Column(
+        Integer,
```
sgallagher, line 174: Comment IDs are actually going to need to be strings. They're unique in Gitlab in that they're a long hex string rather than an integer ID.

logdetective/server/database/models/merge_requests.py
```diff
@@ -0,0 +1,462 @@
...
+    id = Column(Integer, primary_key=True)
+    comment_id = Column(
+        Integer,
+        ForeignKey("comments.id"),
+        nullable=False,
+        index=True,
+        comment="The associated comment id",
+    )
+    reaction_type = Column(
+        String(50),  # e.g. 'thumbs-up'
```
sgallagher, line 337: I went down the rabbit-hole on this, to discover that the current longest emoji supported by Gitlab is `:couple_with_heart_person_person_medium_skin_tone_dark_skin_tone:` which is 66 characters long. We should probably make this string a bit longer. Maybe 100 to be on the safe side?

sgallagher (changes requested):

logdetective/server/database/models/merge_requests.py
abadger:
```suggestion
        String(127),  # e.g. 'thumbs-up'
```

127 or 255 perhaps?

abadger:
The commit https://github.com/fedora-copr/logdetective/pull/217/commits/4034277f5795bca6ae89ce8c2e06f3e8e6b75c24 adding the use of backoff
Fixes #213

majamassarini:
> The commit [4034277](https://github.com/fedora-copr/logdetective/commit/4034277f5795bca6ae89ce8c2e06f3e8e6b75c24) adding the use of backoff Fixes #213

logdetective/server/database/models/merge_requests.py
jpodivin: Yes, but we should create a task about it, noting the tradeoffs.

logdetective/server/database/models/merge_request_jobs.py
```diff
@@ -0,0 +1,473 @@
...
+
+    @classmethod
+    def get_by_details(  # pylint: disable=too-many-arguments disable=too-many-positional-arguments
+        cls,
+        forge: Forge,
+        project_id: int,
+        mr_iid: int,
+        job_id: int,
+        comment_id: str,
+    ) -> Optional["Comments"]:
```
sgallagher, line 234: If we have the `comment_id` and forge, that should be all we need to retrieve this since it's unique and indexed. I think we need two functions here: one that returns all of the comments associated with (forge, project_id, mr_iid) and a function that just returns a single Comment by the `forge_instance` and `comment_id`.

logdetective/server/database/models/merge_request_jobs.py
```diff
@@ -0,0 +1,473 @@
...
+        unique=True,  # 1 comment for 1 job
+        index=True,
+        comment="The associated merge request job (db) id",
+    )
+    comment_id = Column(
+        String(50),  # e.g. 'd5a3ff139356ce33e37e73add446f16869741b50'
+        nullable=False,
+        index=True,
+        comment="The comment gitlab id",
+    )
```
sgallagher, line 178: The `comment_id` is unique within a forge instance. Can you define that constraint, please?

sgallagher (changes requested):
I have a couple remaining changes to request for the Comment objects, but we're almost ready here.

sgallagher (approved):
Looks good!

PR #220, opened 2025-04-15
Always process the logs, just don't comment

sgallagher:
We will probably revert this once we go properly live, but for now we want to be load-testing our setup.


TomasTomecek (approved):
Nice, this was easy!

Let's merge and try it in production tomorrow

@FrostyX can you handle that please?

Issue #208, opened 2025-04-09
Explanation completely misses the point


voxik:
Looking at [log](https://log-detective.com/explain?url=https%3A//download.copr.fedorainfracloud.org/results/vondruch/ror8/fedora-rawhide-
x86_64/08881490-rubygem-actiontext/builder-live.log.gz), the important bits are completely omitted.

Maybe the model is too young. Not sure.

TomasTomecek:
Completely agreed. The problem here is that our processing of logs has failed at detecting the error log line.

We are capturing these kind of logs so we can tune that logic, thank you for sharing it. I added the log here: https://github.com/fedora-copr/logdetective-sample/tree/main/data/0006

jpodivin:
I've performed some experiments using CLI version of the application. It seems that if I increase number of clusters to about 30 I can get reasonable result.

```
 
            Snippet No. 0 at line #88:

            INFO: Reading stdout from command: git rev-parse HEAD

            ================
            
            Snippet No. 1 at line #128:

            Start: clean chroot

            ================
            
            Snippet No. 2 at line #129:

            Finish: clean chroot

            ================
            
            Snippet No. 3 at line #864:

            /bin/tar: Removing leading `/' from member names

            ================
            
            Snippet No. 4 at line #2074:

            + STATUS=0

            ================
            
            Snippet No. 5 at line #2304:

            + BUNDLE_GEMFILE=/builddir/build/BUILD/rubygem-actiontext-8.0.1-build/actiontext-8.0.1/usr/share/gems/gems/actiontext-8.0.1/../Gemfile

            ================
            
            Snippet No. 6 at line #2306:

            + mv test/javascript_package_test.rb test/javascript_package_test.rb.disable

            ================
            
            Snippet No. 7 at line #2307:

            + ruby -Itest -e 'Dir.glob "./test/**/*_test.rb", &method(:require)'

            ================
            
            Snippet No. 8 at line #2308:

            /usr/share/gems/gems/capybara-3.40.0/lib/capybara/session/config.rb:95: warning: URI::RFC3986_PARSER.make_regexp is obsolete. Use URI::RFC2396_PARSER.make_regexp explicitly.

            ================
            
            Snippet No. 9 at line #2309:

            /usr/share/gems/gems/ffi-1.17.0/lib/ffi/function.rb:65: warning: method redefined; discarding old vips_object_print_all

            ================
            
            Snippet No. 10 at line #2311:

            Run options: --seed 30313


            ================
[...]

PR #222, opened 2025-04-16
Increasing number of clusters

jpodivin:
Could help with #208 . The inference will take a bit longer however.


TomasTomecek (approved):

TomasTomecek:
just updated the production config.yml
this will be live with the next update

PR #223, opened 2025-04-16
Create a volume for the matplotlib config and caching area

majamassarini:


TomasTomecek (approved):

FrostyX:
We shouldn't forget to configure it in OpenShift as well.

TomasTomecek:
I created LD-71 for that.

PR #226, opened 2025-04-16
Add a first unit tests for the gitlab endpoint

majamassarini:


tox.ini
```diff
@@ -18,7 +18,8 @@ deps =
     pytest
     pytest-asyncio
     flexmock
-    aioresponses
+    aioresponses  # for testing aiohttp
+    responses  # for testing gitlab which is using requests
```
jpodivin, line 22: Should we also include this in `server-testing` dependencies?

majamassarini: no (the `server-testing` target does exist only to deal with the `psycopg2-binary`, both server and server-testing "targets" in the `pyproject.toml` are runtime dependencies), the libraries needed for the tox testing targets are listed in the `tox.ini` file) .

Tests in testing farm were failing because I forgot to add `python-responses` in the `fmf` files.

PR #230, opened 2025-04-22
makefile: add target server-down

TomasTomecek:


Makefile
```diff
@@ -14,6 +14,9 @@ rebuild-server:
 server-up:
 	$(COMPOSE_ENGINE) -f docker-compose-dev.yaml up --build --force-recreate -d server
 
+server-down:
+	$(COMPOSE_ENGINE) -f docker-compose-dev.yaml up down
```
FrostyX, line 18: Probably a typo? The `up down` doesn't look right

TomasTomecek: oh! thanks for noticing, yes I edited that line poorly

sgallagher (approved):

PR #234, opened 2025-04-23
Catch request exceptions from the inference API

FrostyX:
Fix LD-74


TomasTomecek:
I am carrying this one in #235 hence closing and we'll merge both together

PR #237, opened 2025-04-23
Track metrics for Gitlab-triggered requests

sgallagher:
Gitlab uses the underlying code for staged analysis, but it doesn't call the endpoint directly, meaning it wasn't being saved to the database. This patch adds an argument to the decorator allowing us to override the function name it uses to save to the DB (for backwards compatibility) and moves the decorator to the inner function perform_staged_analysis().

Note: this patch does _not_ implement the linking of the metrics with the Gitlab merge request yet. At the moment, it's only treating it as if it was another external request. I'm still trying to figure out how to populate that relationship.

### DIFF
--- a/logdetective/server/metric.py
+++ b/logdetective/server/metric.py
@@ -50,7 +50,7 @@ def update_metrics(
     )
 
 
-def track_request():
+def track_request(name=None):
     """
     Decorator to track requests metrics
     """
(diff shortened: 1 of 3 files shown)


TomasTomecek (approved):
amazing, what a quick fix! I'll deploy it later

TomasTomecek (approved):
Let's go!

PR #251, opened 2025-05-02
Fix_compressed_log_migration

majamassarini:


FrostyX (approved):
I tested locally and it fixes the issue we are having, thank you very much @majamassarini

PR #252, opened 2025-05-04
Make the docker-compose.yaml file compatible with docker

FrostyX:
These constructs work with podman-compose but docker-compose fails with:

    volumes.matplotlib-config Additional property emptyDir is not allowed
    (root) Additional property volumeMounts is not allowed

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -53,8 +53,3 @@ volumes:
   database_data:
     driver: local
   matplotlib-config:
-    emptyDir: {}
-
-volumeMounts:
-  - name: matplotlib-config
-    mountPath: "${MPLCONFIGDIR}"


docker-compose.yaml
```diff
@@ -53,8 +53,3 @@ volumes:
   database_data:
     driver: local
   matplotlib-config:
-    emptyDir: {}
```
FrostyX, line 56: I guess we can simply remove this?

docker-compose.yaml
```diff
@@ -53,8 +53,3 @@ volumes:
...
-
-volumeMounts:
-  - name: matplotlib-config
-    mountPath: "${MPLCONFIGDIR}"
```
FrostyX, line 60: I think this should be written as 

```
volumes:
  - ${MPLCONFIGDIR}:/something:z
```

but I wasn't sure to which container it should be mounted and to what path.

majamassarini: I have added a new commit with the mapping. Please take a look at it.

majamassarini (approved):

PR #254, opened 2025-05-06
Periodic task for checking emoji feedback

majamassarini:


logdetective/server/emoji.py
```diff
@@ -0,0 +1,53 @@
...
+        for key, value in emoji_counts.items():
+            Reactions.create_or_update(
+                comment.forge,
+                mr_job_db.project_id,
+                mr_job_db.mr_iid,
+                mr_job_db.job_id,
+                comment.comment_id,
+                key,
+                value,
+            )
```
sgallagher, line 53: In order to make this code reusable for the webhook-triggered version, could we refactor this so that we have a separate function that takes a merge request ID, looks up all the comments related to that ID and then processes them?

That way, when we get a webhook trigger from a merge request, I can just call into that function rather than reimplement it.

majamassarini: sure, I will work on it after the standup.

TomasTomecek (approved):

sgallagher (approved):
Looks all good to me. Great work!

logdetective/server/emoji.py
```diff
@@ -0,0 +1,83 @@
...
+    Collect emoji feedback from logdetective comments in the specified MR.
+    Check only MR comments created in the last given period of time.
+    """
+    mr_jobs = GitlabMergeRequestJobs.get_by_mr_iid(gitlab_conn.url, project_id, mr_iid)
+    comments_for_mr = [Comments.get_by_mr_job(mr_job) for mr_job in mr_jobs]
+    comments = [
+        comment
+        for comment in comments_for_mr
+        if comment.created_at.replace(tzinfo=datetime.timezone.utc)
+        > period.get_period_start_time()  # noqa: W503 ruff vs flake
```
sgallagher, line 42: Can you remove the TimePeriod argument here (or make it optional, if you think it's still useful)?

I think if we're making a request on a _specific_ merge request, we should look up all of its comments, even if they're very old. It absolutely makes sense not to update the really old ones when we're iterating through the full set, but if we've been asked to update one specific MR, I think we should check them all.

sgallagher (changes requested):

logdetective/server/emoji.py
sgallagher: We're also missing the Forge argument here, which right now is not _strictly_ needed because we have the connection object, but I'm working on a patch to fix that, because we are accidentally relying on a single Gitlab connection (to one configured Gitlab instance) rather than handling things appropriately for whichever instance has triggered the message. See LD-91 in Jira.

sgallagher (changes requested):
There's a bug here somewhere, because the database still retains a count of 1 for any emoji that has _ever_ been added, even if it was removed.

Testing with another person has showed that it increments and decrements just fine except for the last one that reduces it to zero.

logdetective/server/emoji.py
```diff
@@ -0,0 +1,83 @@
...
+            project = projects[mr_job_db.id]
+        mr_iid = mr_job_db.mr_iid
+        if mr_iid not in mrs:
+            mrs[mr_iid] = mr = await asyncio.to_thread(
+                project.mergerequests.get, mr_iid
+            )
+        else:
+            mr = mrs[mr_iid]
+
+        note = mr.notes.get(comment.comment_id)
```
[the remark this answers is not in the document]
majamassarini: Thanks for testing it in a real scenario, I mocked the db entries and I missed this detail. 
I am wondering if we should save the `note.id` instead of the `discussion.id` [here](https://github.com/fedora-copr/logdetective/blob/80ff3b4708d0cf023583d5f83754b60335fabbc1/logdetective/server/server.py#L788)? In this way we don't need to assume that we are always dealing with the first note, right?

I do the change in any case, we can alway drop the commit...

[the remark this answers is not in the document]
sgallagher: > Thanks for testing it in a real scenario, I mocked the db entries and I missed this detail. I have updated the code with your suggestion still I am wondering if we should save the `note.id` instead of the `discussion.id` [here](https://github.com/fedora-copr/logdetective/blob/80ff3b4708d0cf023583d5f83754b60335fabbc1/logdetective/server/server.py#L788)? In this way we don't need to assume that we are always dealing with the first note, right?

I think you're probably right that it would be better to save the first note ID. I think when I originally wrote things, I didn't realize we could get the note ID back directly from a lookup against the Merge Request.

That said, to switch over will require us to rewrite any existing history in the database, which means writing a script to interrogate Gitlab during the database migration step. Let's open a ticket for that, but we'll discuss it with the rest of the team before we implement.

sgallagher (approved):
Looks like everything is working now. Great job!

PR #261, opened 2025-05-09
Use aiolimiter for rate-control

sgallagher:
Instead of simply delaying requests, let's take advantage of aiolimiter,
which provides an implementation of the "leaky bucket" algorithm to
handle "bursts" of requests.


sgallagher:
If we agree on this direction, I'll package `python-aiolimiter` for Fedora, but right now it's not available there.

logdetective/server/llm.py
```diff
@@ -68,30 +67,31 @@ async def submit_to_llm_endpoint(
...
-        return response
-    try:
-        return json.loads(await response.text())
-    except UnicodeDecodeError as ex:
-        LOG.error("Error encountered while parsing llama server response: %s", ex)
-        raise HTTPException(
-            status_code=400,
-            detail=f"Couldn't parse the response.\nError: {ex}\nData: {response.text}",
-        ) from ex
+    async with SERVER_CONFIG.inference.get_limiter():
```
TomasTomecek, line 70: sweet, this is a more convenient interface then passing the queue

TomasTomecek (commented):
Very straightforward change, really love the new interface. Let's test it and then we can tweak the configuration.

TomasTomecek (approved):
Let's go!

PR #262, opened 2025-05-13
Create graph for emoji feedback

majamassarini:
tmpg0xtmafd
tmp9qpra9r2
tmp6gx1zemn


TomasTomecek (approved):
Stellar work Maja, as always!

I was wondering if we could put the actual utf8 emoji inside the plot labels

PR #263, opened 2025-05-13
Substitute backoff.expo with backoff.constant

majamassarini:
On production server the llm takes 30 seconds on avarage to give a response.
If llm is not responding wait twice the response time, then 4 times and then 10 times.

### DIFF
--- a/logdetective/server/llm.py
+++ b/logdetective/server/llm.py
@@ -115,7 +115,7 @@ def we_give_up(details: backoff._typing.Details):
 
 
 @backoff.on_exception(
-    backoff.expo,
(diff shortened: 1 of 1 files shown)


logdetective/server/llm.py
```diff
@@ -115,7 +115,7 @@ def we_give_up(details: backoff._typing.Details):
 
 
 @backoff.on_exception(
-    backoff.expo,
+    lambda: backoff.constant([10, 30, 120]),
```
TomasTomecek, line 118: okay, that's very creative :D

TomasTomecek (approved):
Tests passing, let's go!

PR #265, opened 2025-05-13
Fix tracking metrics for the analyze/staged endpoint

majamassarini:
I just realized that the metrics for the analyzed/staged endpoint weren't collected properly.


sgallagher (approved):
Thanks! Looks good.

PR #268, opened 2025-05-14
Change backoff.jitter function

majamassarini:
When submitting snippets to llm on retries upon an exception:
- always wait at least the amount of time defined in the wait_gen (jitters should only increase time)
- retries also upon 400 errors
- print last exception


TomasTomecek (approved):
Just tested this locally and works exactly as we need

PR #270, opened 2025-05-14
comment on all RoG MRs

TomasTomecek:


logdetective/server/gitlab.py
```diff
@@ -92,10 +92,12 @@ async def process_gitlab_job_event(
     update_metrics(metrics_id, staged_response)
     preprocessed_log.close()
 
-    # check if this project is on the opt-in list for posting comments.
-    if project.name not in SERVER_CONFIG.general.packages:
-        LOG.info("Not publishing comment for unrecognized package %s", project.name)
-        return
+    # leaving this section here on purpose so we can swap it with a "deny list"
```
sgallagher, line 95: Instead of commenting this section out, let's replace it with a function `check_supported_packages()` that currently just returns `True` on all requests. Then we still have an obvious place to modify later if needed.

sgallagher (changes requested):

logdetective/server/gitlab.py
sgallagher: Actually, in order to avoid needing to push out a fresh build in the future, it would probably be better to implement an opt-out now, so all we need to do is adjust the config file later.

I'll work something up and send a merge request shortly. (I can scavenge some of the implementation from onoe of my other projects.)

TomasTomecek: very good suggestion, will do that

TomasTomecek: hmmm, so github only showed me the first comment, damn

PR #271, opened 2025-05-14
Enhance the opt-in/out handling

sgallagher:
Make opt-in and opt-out regular expression lists, so we can easily update the contents via the config file.


TomasTomecek (approved):
This is excellent!

PR #273, opened 2025-05-14
Return task_failed if no other log identified

sgallagher:


TomasTomecek (approved):
Interesting, very nice catch. Curious what those logs will contain.

PR #277, opened 2025-05-15
Add tests for Config definition

majamassarini:


TomasTomecek (approved):
this is excellent, exactly what we need
thanks Maja

PR #278, opened 2025-05-15
Schedule collect emoji script every hour

majamassarini:


logdetective/server/server.py
```diff
@@ -504,15 +504,9 @@ async def collect_emoji_task():
...
+    """Schedule the collect_emojis_task to run every hour"""
     while True:
-        now = datetime.datetime.now(datetime.timezone.utc)
-        midnight = datetime.datetime.combine(
-            now.date() + datetime.timedelta(days=1),
-            datetime.time(0, 0),
-            datetime.timezone.utc,
-        )
-        seconds_until_run = (midnight - now).total_seconds()
+        seconds_until_run = 60 * 60
```
sgallagher, line 509: This won't run it every hour, it will insert an hour gap between each run. Which is fine, but I'd like it to be configurable at least. When we fix the issue with the webhooks, I'd like to be able to revert this to daily without requiring code changes.

logdetective/server/server.py
```diff
@@ -504,16 +504,9 @@ async def collect_emoji_task():
 
 
 async def schedule_collect_emojis_task():
-    """Schedule the collect_emojis_task to run every day at midnight"""
+    """Schedule the collect_emojis_task to run every hour"""
```
sgallagher: The docstring no longer matches the implementation.

```suggestion
    """Schedule the collect_emojis_task to run on a configured interval"""
```

sgallagher (changes requested):
This will require an update to the tests for the new configuration option, will it not?

sgallagher (approved):

PR #279, opened 2025-05-15
Gitlab: Support shared key authentication

sgallagher:
We will reject any requests to the Gitlab endpoints that do not carry the X-Gitlab-Token header with a known shared key. This will prevent us from processing requests submitted by unknown clients.

### DIFF
--- a/server/config.yml
+++ b/server/config.yml
@@ -17,11 +17,14 @@ gitlab:
   "GitLab SaaS":
     url: https://gitlab.com
     api_token: glpat-XXXXXX
+    webhook_secrets: []
     max_artifact_size: 300
   "GitLab Internal":
     url: https://gitlab.example.com
     api_token: glpat-XXXXXX
     max_artifact_size: 300
+    webhook_secrets:
+      - example_secret
 general:
   devmode: False
   packages:
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -191,6 +191,15 @@ class GitLabInstanceConfig(BaseModel):
     url: str = None
     api_url: str = None
     api_token: str = None
+
+    # This is a list to support key rotation.
+    # When the key is being changed, we will add the new key as a new entry in
+    # the configuration and then remove the old key once all of the client
+    # webhook configurations have been updated.
+    # If this option is left empty or unspecified, all requests will be
+    # considered authorized.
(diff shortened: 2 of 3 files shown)


majamassarini (approved):
This looks good, thanks! I am just wondering if:
- we should put a placeholder in `server/config.yml` file? Is it better to deploy logdetective completely open or completely closed?
- should we create an issue to later secure the `analyze` endpoint? I am not really sure we came up with a solution for that, but as you wrote if logdetective is accessible behind the firewall than these endpoints can be used in a malicious way.

jpodivin (approved):

sgallagher:
> This looks good, thanks! I am just wondering if:
> 
>     * we should put a placeholder in `server/config.yml` file? Is it better to deploy logdetective completely open or completely closed?

I was assuming that the config.yml would be essentially for dev purposes only; it should always be modified for real deployments. (In fact, the config.yml example we ship won't work with Gitlab without manual edits anyway).

>     * should we create an issue to later secure the `analyze` endpoint? I am not really sure we came up with a solution for that, but as you wrote if logdetective is accessible behind the firewall than these endpoints can be used in a malicious way.

Yes, this should be a new ticket because we probably want a slightly different approach here and need to figure out what it will be.

sgallagher:
> I was assuming that the config.yml would be essentially for dev purposes only; it should always be modified for real deployments. (In fact, the config.yml example we ship won't work with Gitlab without manual edits anyway).
> 

I suppose we could put a placeholder in like we do with the Gitlab token...

```
webhook_secrets:
    - example_secret
```

sgallagher:
I made the requested change to the example config and one other modification: I changed to returning 401 instead of 403 if the token is missing or invalid, since that's more in line with the HTTP specification.

server/config.yml
```diff
@@ -17,11 +17,14 @@ gitlab:
...
     url: https://gitlab.com
     api_token: glpat-XXXXXX
+    webhook_secrets: []
     max_artifact_size: 300
   "GitLab Internal":
     url: https://gitlab.example.com
     api_token: glpat-XXXXXX
     max_artifact_size: 300
+    webhook_secrets:
+      - example_secret
```
majamassarini, line 27: Thanks!

PR #280, opened 2025-05-15
Gitlab: Don't process the same build twice

sgallagher:


sgallagher:
The second patch I just pushed removes the catch-and-ignore of an integrity violation that should now be unreachable. If it's not, we want to know about it, I think.

majamassarini (approved):
LGTM, Thanks!

PR #283, opened 2025-05-20
Catch possible Gitlab errors when collecting emojis

majamassarini:


TomasTomecek (approved):
I'm fine with merging this as it is, though I worry that eating the exceptions and just moving on isn't the best strategy. Since you are doing `LOG.error` they should end up in sentry, right? So this is definitely an improvement to the current situation.

majamassarini:
> I'm fine with merging this as it is, though I worry that eating the exceptions and just moving on isn't the best strategy. Since you are doing `LOG.error` they should end up in sentry, right? So this is definitely an improvement to the current situation.

I see your point, I just assumed that the gitlab objects (like a discussion in this case) could be deleted? In this case we should handle the `Not Found` errors, I don't believe we created a non existing discussion id 🤔? And I don't think we should remove the related db data.  But if you have any other suggestion, add it 🙂 It is true I am catching more than just the `Not Found` error, maybe we should retry on  other errors?

sgallagher (changes requested):
So, it's probably _safe_ to merge this, since the emoji collection is periodic and will try again later, but I think I'd really rather that we were more specific with our Exception catching. I think we really do want to know when we're looking up an ID that is missing from the DB. That indicates that we're failing to save things properly.

It's probably safe to catch and ignore a connection-level problem like a timeout or a 500 error, though. We'll just pick that up on the next iteration of the task.

logdetective/server/emoji.py
```diff
@@ -36,6 +37,23 @@ async def collect_emojis_for_mr(
...
+    except gitlab.GitlabError as e:
+        log_msg = f"Error during GitLab operation {func}{args}: {e}"
+        if "Not Found" in str(e):
+            LOG.error(log_msg)
+        else:
+            LOG.info(log_msg)
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        LOG.error("Unexpected error during GitLab operation %s(%s): %s", func, args, e)
+
+
```
sgallagher, line 56:
```suggestion
async def _handle_gitlab_operation(func: Callable, *args):
    """
    It handles errors for the specified GitLab operation.
    After executing it in a separate thread.
    """
    try:
        return await asyncio.to_thread(func, *args)
    except gitlab.GitlabError as e:
        log_msg = f"Error during GitLab operation {func}{args}: {e}"
        if "Not Found" in str(e):
            LOG.error(log_msg)
        else:
            LOG.info(log_msg)


```

At minimum, we need to drop the catch-all statement here; we want to know if something unexpected happens and have it show up in the Sentry report. Since this is happening in a background task, FastAPI will catch it anyway and log it with the appropriate backtrace.

I also don't like how broad the GitlabError catch is, though I'm not certain it'll be realistic to detect all of the probable cases; I'm going to look into narrowing that down.

sgallagher (approved):

PR #286, opened 2025-05-20
DRAFT: Get options from InferenceConfig

sgallagher:
This will allow us to have different models and even servers for different tasks in Log Detective. It also ensures that we have separate rate-limiters for each such object.

Moves the creation and destruction of the HTTP Session into the InferenceConfig object.


sgallagher:
This is still a work-in-progress. Some tests are failing due to making the config reading require async actions and I haven't ensured that ALL uses of the `submit_*()` functions have been updated yet.

I will continue working on this tomorrow, but I'd appreciate a review of the overall idea if possible.

Basically, I am moving away from reading everything from the global `SERVER_CONFIG` variable and instead carrying data in the `inference_cfg` options. This allow us to have different models, max_tokens, API auth tokens, timeouts, rate-limits, etc. for different InferenceConfig instances. It will enable us to have different inference servers for different tasks, including things like having a separate inference server and/or model for each of pre-processing (draining the logs), analysis, and post-processing (safety and sanity).

TomasTomecek (approved):
Really love this new structure, much cleaner function calls, thank you!

sgallagher:
@TomasTomecek Could you have a look at this when you get in tomorrow? I can't figure out how to reproduce the failure that's happening in the Test Farm tests locally. (The `tmt-reproducer` output doesn't work.)

TomasTomecek:
I can reproduce it directly in my environment.
```
FAILED tests/test_server.py::test_submit_to_llm - aiohttp.client_exceptions.ClientConnectionError: Connection refused: POST /v1/chat/completions
FAILED tests/test_server.py::test_submit_text_chat_completions - aiohttp.client_exceptions.ClientConnectionError: Connection refused: POST /v1/chat/completions
```

```
aiodns==3.2.0
aiohappyeyeballs==2.4.6
aiohttp==3.11.13
aiolimiter==1.2.1
aioresponses==0.7.7
aiosignal==1.3.2
```

sgallagher:
OK, the Testing Farm failures are due to a bug in the `aioresponses` library that was fixed in version 0.7.8, but Fedora 41 has 0.7.6 and Fedora 42 has 0.7.7. I've opened https://bugzilla.redhat.com/show_bug.cgi?id=2368020 to get 0.7.8 backported to Fedora 41 and 42.

In the meantime, I'll finish the rest of my testing and then we can get this landed. We'll just have to ignore the Testing Farm errors until that fix gets in.

PR #291, opened 2025-06-02
Refactoring to use OpenAI library

jpodivin:
Refactor:
- `openai` library is now part of our dependencies
- Instead of handcrafted requests, we are using `openai` API calls and verify the results with their data structures.
- `InferenceConfig` now has default API token set to `"None"` since OpenAI library can't handle zero length strings.

Misc:

- `openai` isn't packaged, it has to be installed from pypi. The version is pinned in both container image and in the lockfile, and must be kept in sync if we do an update.
- `submit_to_llm_endpoint` function wasn't used anywhere and so it was removed, together with tests.
- several unused imports were removed

### DIFF
--- a/logdetective/server/config.py
+++ b/logdetective/server/config.py
@@ -1,8 +1,10 @@
 import os
 import logging
 import yaml
+from openai import AsyncOpenAI
+
 from logdetective.utils import load_prompts
-from logdetective.server.models import Config
+from logdetective.server.models import Config, InferenceConfig
 
 
 def load_server_config(path: str | None) -> Config:
@@ -49,6 +51,14 @@ def get_log(config: Config):
     return log
 
 
+def get_openai_api_client(ineference_config: InferenceConfig):
+    """Set up AsyncOpenAI client with default configuration.
+    """
+    return AsyncOpenAI(
+        api_key=ineference_config.api_token,
+        base_url=ineference_config.url)
+
+
 SERVER_CONFIG_PATH = os.environ.get("LOGDETECTIVE_SERVER_CONF", None)
 SERVER_PROMPT_PATH = os.environ.get("LOGDETECTIVE_PROMPTS", None)
 
@@ -56,3 +66,5 @@ SERVER_CONFIG = load_server_config(SERVER_CONFIG_PATH)
 PROMPT_CONFIG = load_prompts(SERVER_PROMPT_PATH)
 
 LOG = get_log(SERVER_CONFIG)
+
+CLIENT = get_openai_api_client(SERVER_CONFIG.inference)
--- a/logdetective/server/llm.py
+++ b/logdetective/server/llm.py
@@ -1,8 +1,7 @@
 import os
 import asyncio
-import json
 import random
-from typing import List, Tuple, Dict, Any, Union
+from typing import List, Tuple, Union
 
 import backoff
 from fastapi import HTTPException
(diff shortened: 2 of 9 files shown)


logdetective/server/llm.py
```diff
@@ -141,48 +88,44 @@ async def submit_text(
...
+            model=inference_cfg.model,
+            temperature=inference_cfg.temperature,
+        )
 
-    if stream:
+    if isinstance(response, AsyncStream):
         return response
+    if not response.choices[0].message.content:
+        LOG.error("No response content recieved from %s", inference_cfg.url)
+        raise RuntimeError()
```
sgallagher, line 120: Do we need to be checking whether Response exists or has `choices` as an attribute? The code is making a lot of assumptions about the return format.

logdetective/server/config.py
```diff
@@ -49,10 +51,20 @@ def get_log(config: Config):
...
+
 SERVER_CONFIG_PATH = os.environ.get("LOGDETECTIVE_SERVER_CONF", None)
 SERVER_PROMPT_PATH = os.environ.get("LOGDETECTIVE_PROMPTS", None)
 
 SERVER_CONFIG = load_server_config(SERVER_CONFIG_PATH)
 PROMPT_CONFIG = load_prompts(SERVER_PROMPT_PATH)
 
 LOG = get_log(SERVER_CONFIG)
+
+CLIENT = get_openai_api_client(SERVER_CONFIG.inference)
```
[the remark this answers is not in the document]
jpodivin: `AsyncOpenAI` ties together a lot of utility functions for building requests. It's remarkably similar in some ways to what we have already implemented, although it does cover more edge cases. 

For example, regarding authentication, it uses api key, to construct request headers, which are then used with every request [3]. Authentication is, effectively, performed during every request. 

[3] https://github.com/openai/openai-python/blob/cca0970798c2735ffb58879ed494e81f80ffc6af/src/openai/_client.py#L270

logdetective/server/llm.py
jpodivin: I'm working with the data structures defined by OpenAI. They are using pydantic, like we are, so we can be reasonably certain which fields are going to exit. Granted, there are ways of abusing pydantic, but you do have to actively try. 

In this case we have just two types of possible response `AsyncStream` and `ChatCompletion`. The first one is largely unused at this time, we don't have client for it and the current architecture of the service doesn't need it. But it could be useful, for example if we get to implement server based inference in our CLI tool. 

The second `ChatCompletion` is pretty simple. [0] We have a list of choices, each with a message, `ChatCompletionMessage` object. That's ensured by pydantic. 

But certainty stops there. `ChatCompletionMessage` has a `content` field, which *may* be a string. But it can be `None`. There are cases when that can happen, and they aren't necessarily easy to figure out [1][2].
It does seem to be related to tool calling, but I would rather err on the side of caution here and assume that it can happen for other reasons. 

Hence the check and rather broad exception raised. When I was testing this PR, I haven't noticed a single case. But if it does happen, we can implement some handling, mitigation or even solution, depending on particulars.

For example, it could be caused by a bug in VLLM runtime, that's only triggered in very particular circumstances. 



[0] https://github.com/openai/openai-python/blob/cca0970798c2735ffb58879ed494e81f80ffc6af/src/openai/types/chat/chat_completion.py#L43
[1] https://community.openai.com/t/chatcompletionmessag-is-returning-content-none-when-finish-reason-stop/583975
[2] https://community.openai.com/t/finish-reason-stop-but-have-a-tool-calls-and-no-content/820316

jpodivin:
Packit failures should be resolved by https://src.fedoraproject.org/rpms/python-logdetective/pull-request/16

sgallagher (approved):

PR #294, opened 2025-06-07
Fixing optional/server dependencies

jpodivin:
We've let our dependencies drift, between the server and CLI packages. This should rectify the situation.

sentry-sdk is now marked as optional and as part of extras groups server and server testing. This means we won't be installing it unless we actually have to.

Fastapi is no longer used in common module, alternative exception from `aiohttp.web` can be used instead.[1]

[1] https://docs.aiohttp.org/en/latest/web_exceptions.html#aiohttp.web.HTTPBadRequest


TomasTomecek (approved):
Thank you, great update! The build is passing well, good to merge :rocket:

PR #295, opened 2025-06-09
Splitting test environments

jpodivin:
Unfortunately I couldn't figure out how to create a single test environment that would trigger two other test environments, other than directly invoking tox commands. Since I didn't like that solution, I've opted to just keep two separated environments instead.


TomasTomecek (approved):
All pass, this is amazing! I love the split to two separate tox envs. No suggestions from my side.

PR #296, opened 2025-06-09
Adding init filest for tests

jpodivin:
tox and pytest don't seem to care, but just to be sure.


TomasTomecek (approved):
right, good catch

PR #300, opened 2025-06-11
Guarding against None return from sqlalchemy

jpodivin:
There is a possibility that th call may return a None. Granted, it would only happen if there was no results at all. But it's something that may happen nonetheless.


majamassarini (approved):
LGTM. Thanks!

PR #304, opened 2025-06-16
New Feature: Koji Integration

sgallagher:
## Overview
This series of patches adds support for integration with the Koji buildsystem.

## Server Configuration
The config.yml for the Log Detective server will need to add a section like:
```
koji:
  analysis_timeout: 15
  max_artifact_size: 300
  instances:
    "fedora":
      xmlrpc_url: https://koji.fedoraproject.org/kojihub
      tokens:
        - example_token
```
It may provide one or more Koji instances in the `instances` dictionary. Each instance must provide at least one `token` for authentication (support for multiple tokens exists to handle periods of credential rotation gracefully).

The `analysis_timeout` is specified in minutes and is there to ensure that an analysis that is interrupted (such as due to an internal error or a service restart) will not prevent a future request from being started, after some time has passed.

## API specification

The API is provided on both HTTP `GET` and `POST` requests to the path `/analysis/koji/{koji_instance}/{task_id}`.

* `{koji_instance}`: A short string matching the name provided in the `instances` field of the configuration file. In production, this will match the name of the Koji instance provided in the `/etc/koji.conf.d` configuration.
* `{task_id}`: The internal Koji task ID for the build task that should be analyzed. Note that if this is a parent task, Log Detective will attempt to identify an appropriate child task on which to perform the analysis. Heuristics are used to pick a single child task when multiple failures may have occurred.

On the `GET` requests, the previously-performed analysis results will be returned, if it exists. If they do not exist, but a request was previously started, it will return an HTTP 202 (Accepted) status code. If no analysis was requested previously, an HTTP 404 (Not Found) status code will be returned. The body of a reply with the analysis results will be a JSON document of the following format:

```json
{
  "task_id": 134754360,
  "log_file_name": "build.log",
  "response": {
    "explanation": {
      "text": "A string with a general explanation of the problem."
      "response_certainty": "A floating-point number of how accurate the LLM thinks its explanation is."
      "logprobs": {}
      "snippets": [
        {
          "text": "The line from the log that the LLM found interesting"
          "explanation": "The LLM's explanation of this line"
          "response_certainty": "A floating-point number of how accurate the LLM thinks its explanation is."
        }
     ]
  }
}
```
The `logprobs` arguments are internal LLM information and can safely be ignored by consumers. The `snippets` contains additional information about the decision-making process the LLM used and may be of some interest. The format of each of these snippets is the same as the 

On the `POST` requests, the results will be returned if they are available as above. Otherwise, an analysis will begin and a 202 (Accepted) status code will be returned. If the optional header `X-Koji-Callback` is passed as a URL, when the analysis completes, an HTTP `POST` request will be sent to the provided URL, containing a JSON payload with a single attribute:

```json
{
  "task_id": 12345
}
```

## Detailed list of changes:
- Identifying the correct architecture and log from a failed `build` task.
- Retrieving the log contents, with a file-size maximum implemented.
- Configuration file extensions for Koji support
- Support for multiple Koji instances (Fedora, CentOS Stream)
- API endpoint accepting a Koji API URL and task ID which returns the Log Detective analysis of the failure.
- Request submission now triggers background analysis and returns `202 (Accepted)` immediately and for any further requests for that task ID until completion. Subsequent requests will always return the cached response.
- Recognizes when a configurable number of minutes have passed since a request for analysis was started and will start a new one if that time has been exceeded and a new query for the result has come in.
- Alembic upgrade path has been written for the new table.
- In addition to line numbers, we return which log is being processed as well as which Task ID (since it might be a child of the originally-submitted one).
- Implements a registration API for notification that the response is finished.
- Provides separate APIs to request existing results if they exist, or trigger new requests if it doesn't.

### DIFF
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -1,4 +1,5 @@
 import asyncio
+from collections import defaultdict
 import datetime
 from logging import BASIC_FORMAT
 from typing import List, Dict, Optional
@@ -343,6 +344,7 @@ class KojiInstanceConfig(BaseModel):
     tokens: List[str] = []
 
     _conn: Optional[koji.ClientSession] = None
+    _callbacks: defaultdict[int, set[str]] = defaultdict(set)
 
     def __init__(self, name: str, data: Optional[dict] = None):
         super().__init__()
@@ -365,6 +367,21 @@ class KojiInstanceConfig(BaseModel):
         if not self._conn:
             self._conn = koji.ClientSession(self.xmlrpc_url)
         return self._conn
+
+    def register_callback(self, task_id: int, callback: str):
+        """Register a callback for a task"""
+        self._callbacks[task_id].add(callback)
+
+    def clear_callbacks(self, task_id: int):
+        """Unregister a callback for a task"""
+        try:
+            del self._callbacks[task_id]
+        except KeyError:
+            pass
+
+    def get_callbacks(self, task_id: int) -> set[str]:
+        """Get the callbacks for a task"""
+        return self._callbacks[task_id]
 
 
 class KojiConfig(BaseModel):
--- a/logdetective/server/server.py
+++ b/logdetective/server/server.py
@@ -259,6 +259,7 @@ async def analyze_rpmbuild_koji(
     koji_instance: Annotated[str, Path(title="The Koji instance to use")],
     task_id: Annotated[int, Path(title="The task ID to analyze")],
     x_koji_token: Annotated[str, Header()] = "",
+    x_koji_callback: Annotated[str, Header()] = "",
     background_tasks: BackgroundTasks = BackgroundTasks(),
 ):
     """Provide endpoint for retrieving log file analysis of a Koji task"""
@@ -284,28 +285,23 @@ async def analyze_rpmbuild_koji(
         # This task ID is malformed or out of range, so we will return a 400.
         response = BasicResponse(status_code=404, content="Invalid or unknown task ID.")
 
-    except KojiTaskNotFoundError:
-        # Task not yet analyzed, so we need to start the analysis in the
-        # background and return a 202 (Accepted) error.
+    except (KojiTaskNotFoundError, KojiTaskAnalysisTimeoutError):
+        # Task not yet analyzed or it timed out, so we need to start the
+        # analysis in the background and return a 202 (Accepted) error.
+
         background_tasks.add_task(
             analyze_koji_task,
             task_id,
             koji_instance_config,
-        )
-        response = BasicResponse(
-            status_code=202, content=f"Analysis begun for task {task_id}"
         )
 
-    except KojiTaskAnalysisTimeoutError:
-        # Task analysis has timed out, so we assume that the request was lost
-        # and that we need to start another analysis.
-        background_tasks.add_task(
-            analyze_koji_task,
-            task_id,
-            koji_instance_config,
-        )
+        # If a callback URL is provided, we need to add it to the callbacks
+        # table so that we can notify it when the analysis is complete.
+        if x_koji_callback:
+            koji_instance_config.register_callback(task_id, x_koji_callback)
+
         response = BasicResponse(
-            status_code=202, content=f"Retrying analysis for task {task_id}"
+            status_code=202, content=f"Beginning analysis of task {task_id}"
         )
 
     except KojiTaskNotAnalyzedError:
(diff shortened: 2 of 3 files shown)


sgallagher:
This is now implemented and working, but it occurred to me today that we still have one problem with it: the turnaround time on the reply is much too long for a single HTTP transaction. I need to do a bit more design, but I think what we need to do is return HTTP 202 (Accepted) immediately (and on subsequent attempts to retrieve information about the same task) and return it only when the results have been processed and stored, either permanently in the database or at least in a temporary cache.

This will require some architecture discussions early next week to hash out the details before I start implementation.

logdetective/server/database/models/koji.py
```diff
@@ -0,0 +1,139 @@
...
+    """Exception raised when a task is not found"""
+
+
+class TaskNotAnalyzedError(Exception):
+    """Exception raised when a task analysis is still in progress"""
+
+
+class TaskAnalysisTimeoutError(Exception):
+    """Exception raised when a task analysis has timed out"""
+
```
jpodivin, line 24: wdyt about putting all exceptions into a separate module?

logdetective/server/metric.py
```diff
@@ -17,8 +17,8 @@
 
 async def add_new_metrics(
     api_name: str,
-    url: str,
-    http_session: aiohttp.ClientSession,
+    url: str = None,
+    http_session: aiohttp.ClientSession = None,
```
jpodivin, line 21: We'll need to change typing hints to `Optional` for this. Otherwise the IDE will complain. Also, the same change should be applied to other arguments.

alembic/versions/ae776dc8160b_add_koji_task_analysis_table.py
```diff
@@ -0,0 +1,121 @@
...
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column(
+            "koji_instance",
+            sa.Enum(
+                "https://brewhub.engineering.redhat.com/brewhub",
+                "https://cbs.centos.org/kojihub/",
+                "https://koji.fedoraproject.org/kojihub",
+                "https://s390.koji.fedoraproject.org/kojihub",
+                "https://koji.stg.fedoraproject.org/kojihub",
+                "https://kojihub.stream.rdu2.redhat.com/kojihub",
```
jpodivin, line 48: Would it make sense to set these values in a separate variable and use it both cases?

logdetective/server/models.py
```diff
@@ -323,6 +331,63 @@ def __init__(self, data: Optional[dict] = None):
...
+            return
+
+        if "analysis_timeout" in data:
+            self.analysis_timeout = data["analysis_timeout"]
+            del data["analysis_timeout"]
+        else:
+            self.analysis_timeout = 15
+
+        for instance_name, instance_data in data.items():
+            self.instances[instance_name] = KojiInstanceConfig(
```
jpodivin, line 386: We could avoid this by putting configuration for individual instances into a list of its own, accessible by `koji_instances` key. That way the conditional and `del` on line 381 wouldn't be necessary.

logdetective/server/models.py
```diff
@@ -323,6 +331,63 @@ def __init__(self, data: Optional[dict] = None):
...
+            # Set some reasonable defaults
+            self.xmlrpc_url = "https://koji.fedoraproject.org/kojihub"
+            self.tokens = []
+            self.max_artifact_size = 1024 * 1024
+            return
+
+        self.xmlrpc_url = data.get(
+            "xmlrpc_url", "https://koji.fedoraproject.org/kojihub"
+        )
+        self.max_artifact_size = int(data.get("max_artifact_size")) * 1024 * 1024
```
jpodivin, line 358: Default should be set to some reasonable value, probably `300` as it is on 339

jpodivin (commented):
Couple of suggestions but nothing too important.

logdetective/server/database/models/koji.py
sgallagher: I'm not opposed, but as they're somewhat generically-named, I figured keeping them in the `logdetective.server.koji` hierarchy meant that they'd be unlikely to get confused for other tasks.

If we move them, I'll probably rename them to be `KojiTask...` for clarity.

alembic/versions/ae776dc8160b_add_koji_task_analysis_table.py
sgallagher: Yeah, we can simplify this. I had this code generated by alembic and didn't realize it duplicated that.

logdetective/server/models.py
sgallagher: That's a typo. It was indeed supposed to be 300 there.

logdetective/server/metric.py
jpodivin: Indeed. It's something we should do, but it's somewhat lower priority than... a lot of stuff. It's just that I wouldn't like to introduce more of these cases.

logdetective/server/server.py
```diff
@@ -8,13 +8,30 @@
...
 from fastapi.responses import StreamingResponse
 from fastapi.responses import Response as BasicResponse
 import aiohttp
 import sentry_sdk
 
+from logdetective.server.exceptions import KojiInvalidTaskID
+
+from logdetective.server.database.models.koji import (
+    KojiTaskAnalysis,
+    KojiTaskAnalysisTimeoutError,
```
jpodivin, line 30: I don't think these go against any guidelines, not that I care much about those, but for clarity it may be better to import them from module they are defined in.

sgallagher: Yeah, those imports were generated by my IDE and I didn't check to make sure they were imported from the primary source. I'll adjust it.

alembic/versions/ae776dc8160b_add_koji_task_analysis_table.py
sgallagher: This actually got completely replaced in the latest changes, as I realized that we need the set of Koji instances to be variable, since there are some out there -- like Datto -- whose [internal] URLs I don't know and therefore can't hard-code.

In the latest iteration, this is set in the config.yml and saved to the Database.

PR #320, opened 2025-07-02
Allow for explicit filtering of log patterns

jpodivin:
Some log lines are just not useful at all. We can filter them out explicitly, by providing a dictionary of matching regular expressions. 

The `skip_snippets.yml` holds a dictionary of patterns we want to avoid when processing logs. Each entry has a key, which should denote what the pattern is meant to match, thus documenting the intention and also providing useful information in our logs.

The regular expressions are compiled ahead, to save time, and are then matched against chunks before they are submitted to drain. Matching chunks are skipped. The test is executed twice, because there is a chance that a template generated by drain may match one of the excluded patterns. This behavior can occur when a very similar text is present in the logs. Even though it didn't match the specified regular expression, the template generated from it can, thanks to masking.

Feature is enabled in both CLI and server applications.

Empty file will simply allow all chunks to be processed, maintaining existing behavior.

Special care needs to be taken when writing regular expressions in the file. Certain patterns may match syntax of yaml, and be processed by pydanting as part of the data structure.

### DIFF
--- a/.env
+++ b/.env
@@ -28,6 +28,8 @@ LOGDETECTIVE_SERVER_CONF="/config.yml"
 LOGDETECTIVE_PROMPTS="/src/logdetective/prompts.yml"
 # llama.cpp inference server will look for `LLAMA_ARG_MODEL` in this location
 MODELS_PATH="./models"
+# Path to patterns to be skipped during processing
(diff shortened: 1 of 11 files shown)


sgallagher (approved):
Looks good to me now. Thanks!

PR #324, opened 2025-07-09
Switching default to a more advanced model

jpodivin:


sgallagher (changes requested):
Please update https://github.com/fedora-copr/logdetective/tree/main?tab=readme-ov-file#server with instructions to pull down the new default model.

sgallagher (approved):

PR #325, opened 2025-07-10
Adding guidelines for submission of potentiall harmful content.

jpodivin:
While we are only using secure models in deployment, and our prompt set up to guide model towards appropriate output, there is still a chance of our system being subverted. One of the most simple options would be prompt injection attack, with well crafted log submission.

While there are options for further mitigation, it is necessary to acknowledge that complete assurance is not possible. 
Therefore we will direct users of our gitlab service to submit formal reports in Gitlab, using existing functionality.

At the same time we will encourage them to contact us.

### DIFF
--- a/logdetective/server/templates/gitlab_full_comment.md.j2
+++ b/logdetective/server/templates/gitlab_full_comment.md.j2
@@ -59,7 +59,9 @@ Was the provided feedback accurate and helpful? <br>Please vote with :thumbsup:
 or :thumbsdown: to help us improve.<br>
 
 <i>Should Log Detective report contain harmful content, please use Gitlab reporting feature
-at the top right corner of the comment box and contact developers.</i>
+<i>If this Log Detective report contains harmful content, please use the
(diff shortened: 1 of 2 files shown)


sgallagher (changes requested):

logdetective/server/templates/gitlab_full_comment.md.j2
```diff
@@ -58,7 +58,10 @@ This comment was created by [Log Detective][log-detective].
 Was the provided feedback accurate and helpful? <br>Please vote with :thumbsup:
 or :thumbsdown: to help us improve.<br>
 
-
+<i>Should Log Detective report contain harmful content, please use Gitlab reporting feature
```
sgallagher, line 61: I think this line is extra.

sgallagher (changes requested):

logdetective/server/templates/gitlab_full_comment.md.j2
jpodivin: Interesting. I've applied the patches exactly as proposed, I suppose github couldn't apply them properly.

PR #327, opened 2025-07-14
Create note in MR rather than a discussion

ckelleyRH:
Unresolved threads in GitLab MRs block merges In CS/RHEL development. The MRs also require approval, which is not provided by automation in the case of failed builds, so it is not necessary to force the author to acknowledge the thread.

In [this example](https://gitlab.com/redhat/centos-stream/rpms/selinux-policy/-/merge_requests/208), the author had the builds fail then succeed in a subsequent run. They got the necessary approvals, and resolved the failed results thread, but failed to notice that Log Detective had also opened a thread.

I think a note, which is non-blocking, is sufficient but it is possible I am missing supporting context and a discussion is in fact necessary.

### DIFF
--- a/logdetective/server/gitlab.py
+++ b/logdetective/server/gitlab.py
@@ -329,7 +329,7 @@ async def comment_on_mr(  # pylint: disable=too-many-arguments disable=too-many-
 
     # Submit a new comment to the Merge Request using the Gitlab API
     discussion = await asyncio.to_thread(
-        merge_request.discussions.create, {"body": short_comment}
+        merge_request.notes.create, {"body": short_comment}
     )
 
     # Get the ID of the first note


sgallagher (changes requested):
You're correct that we probably do not need to be starting a discussion and instead could be posting a note; that's an oversight on my part. (I think I was following a documentation example and didn't realize the difference inherent in starting a Discussion there).

However, this patch is nowhere near complete if we want to change this. Among other things, we actually store the Discussion ID in our database and use that to look up and suppress the earlier messages if and when a subsequent Log Detective analysis is run (such as when a new set of patches are pushed and the build fails again).

To solve this properly, we need to update all uses of Discussions to use Notes instead, which means also performing a database migration to convert existing entries to Note IDs instead of Discussion IDs.

ckelleyRH:
> You're correct that we probably do not need to be starting a discussion and instead could be posting a note; that's an oversight on my part. (I think I was following a documentation example and didn't realize the difference inherent in starting a Discussion there).
> 
> However, this patch is nowhere near complete if we want to change this. Among other things, we actually store the Discussion ID in our database and use that to look up and suppress the earlier messages if and when a subsequent Log Detective analysis is run (such as when a new set of patches are pushed and the build fails again).
> 
> To solve this properly, we need to update all uses of Discussions to use Notes instead, which means also performing a database migration to convert existing entries to Note IDs instead of Discussion IDs.

I see, I thanks for the explanation. As this is much more involved than it first appeared I raised an issue for it: https://github.com/fedora-copr/logdetective/issues/331

jpodivin:
Closing. Similar PR may be opened as result of #331

PR #333, opened 2025-07-16
Database fixes

jpodivin:


sgallagher (approved):

logdetective/server/database/models/metrics.py
```diff
@@ -204,11 +208,12 @@ def get_postgres_time_format(cls, time_format):
...
-        cls, time_format: str, values_dict: dict[str, any]
-    ) -> dict[datetime.datetime, any]:
-        """Convert from a dictionary with str keys to a dictionary with datetime keys"""
+        cls, time_format: str, query_results: List[Tuple[str, int]]
+    ) -> dict[datetime.datetime, int]:
+        """Convert from a list of tuples with str as first values
+        to a dictionary with datetime keys"""
         new_dict = {
-            datetime.datetime.strptime(r[0], time_format): r[1] for r in values_dict
+            datetime.datetime.strptime(e[0], time_format): e[1] for e in query_results
```
majamassarini, line 216:
```suggestion
            datetime.datetime.strptime(e[0], time_format): e[1] for e in query_results
```
If the row has more than two elements this code will raise an exception, probably the previous line of code was a bit safer. 
But since we have the typing checking for a two elements tuple, I think it is ok also in this way.

majamassarini (approved):
LGTM, thanks!

I have just a small note on the last line.

logdetective/server/database/models/metrics.py
majamassarini: I think it has changed in the meantime 😅. Sorry for the noise

PR #336, opened 2025-07-17
Fixing previous version hash and setting date

jpodivin:
Existing code causes `Multiple head revisions are present for given argument 'head'; please specify a specific target revision, '<branchname>@head' to narrow to a specific head, or 'heads' for all heads` error.


alembic/versions/ae776dc8160b_add_koji_task_analysis_table.py
```diff
@@ -14,7 +14,7 @@
 
 # revision identifiers, used by Alembic.
 revision: str = "ae776dc8160b"
-down_revision: Union[str, None] = "2f60bb92411f"
+down_revision: Union[str, None] = "360aea89443f"
```
majamassarini, line 17:
```suggestion
down_revision: Union[str, None] = "d90c3ebde7ed"
```
I would have expected it to revise this version, but maybe you know something more that I don't.

majamassarini: And if you have tested it locally and is working I am fine with the old value!

jpodivin: The original patch was wrong. The field and comment should always have been `360aea89443f`

PR #341, opened 2025-07-28
Mark comments with missing parent discussion and do not retrieve them…

jpodivin:
… again


sgallagher:
This needs more explanation. The parent discussion shouldn't be 404 ever... which suggests that we might be storing the wrong value.

I wonder if the discussion ID changes if the comment is modified...?

sgallagher:
> In my opinion there are basically two options.
> 
>     1. We are somehow manufacturing discussion ids and storing them in database. These discussions never existed.
> 
>     2. These discussions used to exist, but they were deleted or overwritten. I do know for sure that the MRs do exist, and that they have other comments from Log Detective, so that is a start.
3. We're saving a comment ID in the database and later trying to retrieve it as a discussion ID (or vice versa).

jpodivin:
> > In my opinion there are basically two options.
> > ```
> > 1. We are somehow manufacturing discussion ids and storing them in database. These discussions never existed.
> > 
> > 2. These discussions used to exist, but they were deleted or overwritten. I do know for sure that the MRs do exist, and that they have other comments from Log Detective, so that is a start.
> > ```
> 
>     3. We're saving a comment ID in the database and later trying to retrieve it as a discussion ID (or vice versa).

I have considered that option. But I do not find it likely.

Gitlab API does have a note[1] object, effectively a comment, as a subordinate member of the discussion[2] data structure. Both of these have integer IDs, and documentation doesn't specify more. But when you inspect actual objects, you note that ids of discussions are relatively many digit hexadecimals, where as notes are relatively fewer digit decimals.

This is actually a problem on gitlab side. Their documentation omits the fact that discussion ID is a hash. [3]

The errors we are picking up, all involve attempted lookup with a long hexadecimal. That would indicate that we are indeed dealing with a discussion id, not a note id.

I have checked some of the affected merge requests previously, to see if the id we are looking for isn't present but attached to some other object. But that isn't the case. One possibility that remains, is that we are using correct id, that is id of actual discussion, but attaching it to a wrong MR or wrong project. However when inspecting code adding new discussions (comments) into database, and I couldn't find anything that would point towards that. 

[1] https://docs.gitlab.com/api/notes/
[2] https://docs.gitlab.com/api/discussions/
[3] https://gitlab.com/gitlab-org/gitlab/-/issues/358951

PR #343, opened 2025-08-01
Reworking test dependencies

jpodivin:
All of our test dependencies are now tracked in pyproject.toml and lockfile.


jpodivin:
Unfortunately there is no constraint on the llama.cpp version in downstream. So this won't work there :/

jpodivin:
Tests on rawhide are failing due to lack of support for python3.14 in pydantic https://github.com/pydantic/pydantic/issues/11613 Hopefully this should be resolved soon. In the meantime we have to contend with this.

PR #346, opened 2025-08-05
Limit on maximum length of a snippet

jpodivin:
Template mining is generally reliable when it comes to extracting significant snippets of build logs and limiting the amount of text we send to LLMs. Unfortunately, there are rare edge cases, when our heuristics for chunking produce snippets that are unreasonably large.

When submitted, these snippets cause overflow of model context and failure of analysis. Cases like this are, thankfully, very rare. But we need to guard against them.

This check on maximum number of characters in a snippet takes the pessimistic assumption about tokenizer performance `1 char == 1 token`. Under most circumstances we can expect one token to represent more than one character on average. Nevertheless, in interest of stability and simplicity, the conditional works with characters.

The limit on snippet size must be set in accordance with properties of model used for inference. Setting the limit too high, will, for obvious reasons, effectively prevent the conditional from doing it's job.

### DIFF
--- a/logdetective/extractors.py
+++ b/logdetective/extractors.py
@@ -20,7 +20,8 @@ class DrainExtractor:
         context: bool = False,
         max_clusters=8,
         skip_snippets: SkipSnippets = SkipSnippets({}),
-    ):
+        max_snippet_len: int = 2000
+    ):  # pylint: disable=R0913,R0917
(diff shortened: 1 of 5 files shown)


TomasTomecek (approved):
wow, didn't know they can be so big
very nice change +1

PR #351, opened 2025-08-13
Processing logs with csgrep

jpodivin:
We can use csgrep as a more precise way to extract info from logs. Downside is that rules are more strict, so unless the messages are formatted properly, they won't get into output of csgrep. For that reason, I believe it would be better to use it in addition to drain, not as a replacement.

The new `CSGrepExtractor` is derived from the original `DrainExtractor`, using drain algorithm to extract useful lines.
However, instead of using simple heuristic to pick out chunks of text for clustering, it uses csgrep.
Drain is then used to get rid of duplicates, and to make sure that we don't inundate LLM with requests.
At the same time, since files we are passing to csgrep are not just compiler logs, the csgrep produces a great deal of syntax errors. Thankfully, these don't stop the process of message extraction, even though they change the exit code.

With this change, we are adding an optional dependency on csgrep. The current Log Detective behavior is maintained however, and the application remains backwards compatible. To keep this backwards compatibility, some changes were not implemented and are left for the future. Most notable of these changes is extension of config file `extractor` section, into `extractors` section, with entries for both `CSGrepExtractor` and `DrainExtractor`. 

There are also other, smaller, changes to facilitate work with the new feature:

* Compression ratio is now correctly calculated, since we have moved pass line based extraction
* Extraction of snippets in CLI and server applications are now much closer
* Delimiter between snippets has been turned into a variable, so we don't have to change it in multiple places
* New tests for extractors and for data models used by the servers

### DIFF
--- a/logdetective/extractors.py
+++ b/logdetective/extractors.py
@@ -15,6 +15,7 @@ LOG = logging.getLogger("logdetective")
 
 class Extractor:
     """Base extractor class."""
+
     def __init__(
         self,
         verbose: bool = False,
@@ -28,7 +29,9 @@ class Extractor:
     def __call__(self, log: str) -> list[Tuple[int, str]]:
         raise NotImplementedError
 
-    def filter_snippet_patterns(self, chunks: list[tuple[int, str]]) -> list[tuple[int, str]]:
+    def filter_snippet_patterns(
+        self, chunks: list[tuple[int, str]]
+    ) -> list[tuple[int, str]]:
         """Keep only chunks that don't match any of the excluded patterns"""
         chunks = [
             (_, chunk)
@@ -114,7 +117,13 @@ class CSGrepExtractor(DrainExtractor):
             # We are not running binary in check mode, since csgrep
             # can produce many errors due to log file syntax
             result = sp.run(
-                ["csgrep", "--event=error", "--remove-duplicates", "--mode=json"],
+                [
+                    "csgrep",
+                    "--event=error",
+                    "--remove-duplicates",
+                    "--mode=json",
+                    "--quiet",
+                ],
                 input=log,
                 shell=False,
                 check=False,


jpodivin:
> Wow, fascinating; would love to see a sample output for a failure in a C package. Let me know when this is deployed :)

When using one of our collected samples https://github.com/fedora-copr/logdetective-sample/tree/ad0986f162f97243eff5407dee7d08ad7d0f1100/data/b4a1a122-055f-472d-a0be-0a18dbbf2543 with CLI Logdetective, the results look like this:

```
INFO:logdetective:Log summary: 
 
            Snippet No. 0 at line #6924:

            ================
            
            Snippet No. 1 at line #6925:

              File "/usr/lib/python3.13/site-packages/mockbuild/trace_decorator.py", line 93, in trace

            ================
            
            Snippet No. 2 at line #6926:

                result = func(*args, **kw)

            ================
            
            Snippet No. 3 at line #6927:

              File "/usr/lib/python3.13/site-packages/mockbuild/util.py", line 610, in do_with_status

            ================
            
            Snippet No. 4 at line #6928:

                raise exception.Error("Command failed: \n # %s\n%s" % (cmd_pretty(command, env), output), child.returncode)

            ================
            
            Snippet No. 5 at line #6929:

            mockbuild.exception.Error: Command failed: 

            ================
            
            Snippet No. 6 at line #6930:

             # /usr/bin/systemd-nspawn -q -M 05f5495e64854228a376902df51cc633 -D /var/lib/mock/f41-build-side-104731-57016776-6547491/root -a -u mockbuild --capability=cap_ipc_lock --bind=/tmp/mock-resolv.d3k12dr_:/etc/resolv.conf --bind=/dev/btrfs-control --bind=/dev/mapper/control --bind=/dev/fuse --bind=/dev/loop-control --bind=/dev/loop0 --bind=/dev/loop1 --bind=/dev/loop2 --bind=/dev/loop3 --bind=/dev/loop4 --bind=/dev/loop5 --bind=/dev/loop6 --bind=/dev/loop7 --bind=/dev/loop8 --bind=/dev/loop9 --bind=/dev/loop10 --bind=/dev/loop11 --console=pipe --setenv=TERM=vt100 --setenv=SHELL=/bin/bash --setenv=HOME=/builddir --setenv=HOSTNAME=mock --setenv=PATH=/usr/bin:/bin:/usr/sbin:/sbin '--setenv=PROMPT_COMMAND=printf "\033]0;<mock-chroot>\007"' '--setenv=PS1=<mock-chroot> \s-\v\$ ' --setenv=LANG=C.UTF-8 --resolv-conf=off bash --login -c '/usr/bin/rpmbuild -bb --noclean --target aarch64 --nodeps /builddir/build/SPECS/LabPlot.spec'

            ================
            
            Snippet No. 7 at line #6931:

            

            ================
            
            Snippet No. 8 at line #0:

            In member function ‘bool XLSXFilterPrivate::dataRangeCanBeExportedToMatrix(const QXlsx::CellRange&) const’
unable to deduce ‘const auto*’ from ‘QXlsx::Document::cellAt(int, int) const(i, j)’
  760 |                         const auto* cell = m_document->cellAt(i, j);
      |                                            ~~~~~~~~~~~~~~~~~~^~~~~~
  mismatched types ‘const auto*’ and ‘std::shared_ptr<QXlsx::Cell>’
  760 |                         const auto* cell = m_document->cellAt(i, j);
      |                                                                   ^
            ================
            
            Snippet No. 9 at line #0:

            In member function ‘QXlsx::Cell::CellType XLSXFilterPrivate::columnTypeInRange(int, const QXlsx::CellRange&) const’
unable to deduce ‘const auto*’ from ‘QXlsx::Document::cellAt(int, int) const(row, ((int)column))’
  795 |                         const auto* cell = m_document->cellAt(row, column);
      |                                            ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
  mismatched types ‘const auto*’ and ‘std::shared_ptr<QXlsx::Cell>’
  795 |                         const auto* cell = m_document->cellAt(row, column);
      |                                                                          ^
            ================
            
Explanation:
- [Snippet No. 0 at line #6924 to #6929] : The package build failed due to a command execution error. The command in question is a RPM build command executed within a systemd-nspawn container.
- [Snippet No. 8 at line #0] : An issue with the C++ code in LabPlot.spec file. The compiler is unable to deduce 'const auto*' from 'QXlsx::Document::cellAt(int, int) const(i, j)' which is causing a mismatched types error.
- [Snippet No. 9 at line #0] : Similar issue as Snippet No. 8, but with the function 'QXLSXFilterPrivate::columnTypeInRange(int, const QXlsx::CellRange&) const'. The compiler is unable to deduce 'const auto*' from 'QXlsx::Document::cellAt(int, int) const(row, ((int)column))'.

Complete Explanation and Recommendation:
The RPM build process for the LabPlot package encountered two separate issues. The primary issue occurred during the build command execution, causing it to fail. To resolve this issue, it is crucial to identify the reason for command failure within the systemd-nspawn container and address it.

Secondly, there are two C++ code errors in the LabPlot.spec file related to the use of 'QXlsx::Document::cellAt(int, int) const' function. The compiler is unable to deduce 'const auto*' from the function's argument. To resolve these issues, you should consult the Qt XLSX library documentation or seek assistance from developers experienced with this library to understand the correct way to use these functions. It is likely that the necessary adjustments can be made in the LabPlot.spec file to correct these type mismatches. After resolving these issues, you can rebuild the RPM package.
Response certainty: 95.56%
```

kdudka:
> At the same time, since files we are passing to csgrep are not just compiler logs, the csgrep produces a great deal of
> syntax errors. Thankfully, these don't stop the process of message extraction, even though they change the exit code.

This behavior can be suppressed by the `--ignore-parser-warnings` option of `csgrep`.  As you say, parser warnings are totally expected in your case.  Still the exit code can signal other failures that are not expected, and which should be handled if they occur.  Similarly you may consider the `--quiet` option to suppress the parser warnings in `stderr` so that unexpected errors are easier to spot in the error log.

jpodivin:
> `ignore-parser-warnings`

Thank you, that makes a lot of sense.

PR #375, opened 2025-10-21
Refactoring database to use asynchronous calls

jpodivin:
Chance of something going wrong here are significant, given the extend of changes in code. I would recommend testing thoroughly. 

The database is now working with asyncpg as a backend.
Extensive changes to most of the API were needed, as well as to the ORM. The asyncpg is more strict about types. This requires modifications of the types used, as well as casting.

The API used to perform queries over the database had to be replaced with sqalchemy 2.0 equivalent. Resulting code is more verbose, but performs the same operations.

The PR does include ruff formatting of the affected files.

### DIFF
--- a/tests/server/test_database_models_metrics.py
+++ b/tests/server/test_database_models_metrics.py
@@ -113,7 +113,7 @@ async def test_AnalyzeRequestMetrics_ger_responses_average_length(endpoint):
         end_time = datetime.datetime.now(datetime.timezone.utc)
         start_time = end_time - datetime.timedelta(hours=10)
         time_format = "%Y-%m-%d %H"
-        average_lengths_dict =  (
+        average_lengths_dict = (
             await AnalyzeRequestMetrics.get_responses_average_length_in_period(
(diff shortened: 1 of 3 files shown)


tests/server/test_database_models_metrics.py
```diff
@@ -97,16 +103,17 @@ def test_AnalyzeRequestMetrics_ger_responses_average_time(endpoint):
...
+    async with PopulateDatabase.populate_db(
         duration=duration,
         endpoint=endpoint,
     ) as _:
         end_time = datetime.datetime.now(datetime.timezone.utc)
         start_time = end_time - datetime.timedelta(hours=10)
         time_format = "%Y-%m-%d %H"
-        average_lengths_dict = (
+        average_lengths_dict = await (
             AnalyzeRequestMetrics.get_responses_average_length_in_period(
```
majamassarini:
```suggestion
        average_lengths_dict =  (
            await AnalyzeRequestMetrics.get_responses_average_length_in_period(
```

The same as above.

PR #446, opened 2026-02-12
Reworking visibility of certainty ratings 

jpodivin:
Certainty ratings in CLI will now be on `debug` level of the output, and visible only if the verbosity is increased.

Certainty ratings in server mode, will only be outputted, if `report_certainty` field in config is set to `True`. The value is set to `False` by default. If the `report_certainty` field is set to false, the certainty will not be rendered in the gitlab comment, even if it has been calculated.

### DIFF
--- a/logdetective/server/templates/gitlab_full_comment.md.j2
+++ b/logdetective/server/templates/gitlab_full_comment.md.j2
@@ -1,7 +1,9 @@
 The package {{ package }} build has experienced an issue.
 Please know that the explanation was provided by AI and may be incorrect.
+{% if report_certainty %}
 {% if certainty > 0 %}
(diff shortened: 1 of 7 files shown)


logdetective/server/templates/gitlab_full_comment.md.j2
```diff
@@ -1,8 +1,10 @@
 The package {{ package }} build has experienced an issue.
 Please know that the explanation was provided by AI and may be incorrect.
+{% if report_certainty %}
```
lachmanfrantisek: I see you are still prepared for re-adding... 😁

lachmanfrantisek (approved):

PR #486, opened 2026-04-09
Agentic Log Detective

jpodivin:
This is a complete overhaul of our server and of the way Log Detective works. Instead of using quasi-agentic approach we had before, we are now moving towards fully agentic workflow.

With this change, we will not be able to make release in Fedora until all dependencies are packaged.

- API now accepts multiple build artifacts, both as URLs and as raw strings
- Log Detective is performing analysis by choosing build artifacts and applying extractors to get snippets
- The `/analyze/stream` API endpoint has been removed completely
- The `/analyze/staged` API endpoint has been removed completely
- The `/queue/print` API endpoint has been removed completely
- Log Detective now depends on beeai-framework and indirectly on litellm
- Total size of all artifacts that can be submitted has been set to 50 MB
- Separate inference config for snippets has been removed
- Snippet relevance and sorting functionality has been removed.

Other changes:
- The skip snippet patterns are now using defaults in the same was other configuration files.
- Image for llama.cpp local inference is now based on 7f2cbd9a4df77a2ce10f31d69f07d4fb75eabc07 of the llama.cpp repo
- Dependencies in server image are now almost all pulled from pypi, this has been done to resolve compatibility issues
- Drain3 is now referenced by commit, instead of Pypi release, because maintainer still hasn't released a fixed version
- Default temperature has been set to `0.0` for greater determinism
- Log probabilities handling has been removed
- Legacy response certainty in gitlab analysis was replaced with default smiley face :)


logdetective/prompts/system_prompt.j2
```diff
@@ -6,7 +6,8 @@ You are truthful, concise, and helpful.
 
 ## Input processing
 
-You will work with snippets of logs produced during a failed package build.
+You will work with build artifact produced during a failed RPM build.
```
Jany26: shouldn't it be build "artifacts"?

logdetective/server/agent/agent.py
```diff
@@ -0,0 +1,103 @@
...
+        ConditionalRequirement(
+            ThinkTool, force_at_step=1, consecutive_allowed=False, max_invocations=2
+        ),
+        # At most, allow one invocation of extractor per artifact
+        ConditionalRequirement(
+            DrainExtractorTool,
+            consecutive_allowed=True,
+            min_invocations=1,
+            max_invocations=len(artifacts),
+        ),
```
Jany26, line 49: If I understand this correctly, we allow n uses of the extractor tool for n files, but there is no explicit per-file extraction check... isn't it theoretically possible that an extractor can run multiple times on the same file (and run 0 times on some other file)? Maybe the pre-extraction check is not really needed in practice.

logdetective/server/server.py
```diff
@@ -278,65 +246,31 @@ def requires_token_when_set(authorization: Annotated[str | None, Header()] = Non
...
+        payload, http_session, request_size=request_size
+    )
 
-    return await perform_staged_analysis(
-        log_text,
-        async_request_limiter=request.app.state.llm_request_limiter,
-        extractors=request.app.state.extractors,
+    return await analyze_artifacts(
+        artifacts=artifacts, chat_model=request.app.state.openai_chat_model
     )
```
Jany26, line 268: No sanitization for /analyze?

logdetective/server/models.py
```diff
@@ -128,98 +137,96 @@ class EmojiHook(BaseModel):
...
 
-class Response(BaseModel):
-    """Model of data returned by Log Detective API
+class AnalyzedSnippet(Snippet):
+    """Model for snippets already processed by Log Detective.
 
-    explanation: Explanation
-    response_certainty: float
+    snippet_analysis: LLM output in form of a dictionary
     """
```
Jany26, line 176: Is this a part of per-snippet LLM relevance check? Shouldn't this be no longer supported? I might be missing something.

Jany26 (approved):

logdetective/server/models.py
jpodivin: No, this is something for future development.

logdetective/server/server.py
jpodivin: Now it is is done after pulling each artifact.

logdetective/prompts/system_prompt.j2
jpodivin: Done

logdetective/server/agent/agent.py
jpodivin: I did try something of the sort. But the agent then made too many failed calls. There are probably other ways to do it however. Right now, I'm using a softer approach. Each response has `remaining_artifacts` field, with set of artifacts that were not analyzed yet.

This steers the agent to make it less likely it will choose the same artifact more than once.

The worst case scenario, that the agent will get in the loop, extracting snippets from the same artifact over and over again, running out of attempts and finally returning nonsense, never occurred during my testing.

So I think we may be reasonably safe.

logdetective/server/compressors.py
```diff
@@ -85,7 +84,7 @@ def zip_response(self) -> bytes:
             self.EXPLANATION_FILE_NAME: self._response.explanation.model_dump_json()
         }
 
-        if isinstance(self._response, StagedResponse):
+        if self._response.snippets:
```
Jany26, line 87: this if-condition seems redundant now

jpodivin: Not entirely, the model still allows for the snippets to be `None`. Now this shouldn't happen, but it is technically possible. I would leave this check as it is.

PR #505, opened 2026-05-11
Simplification of prompt management

jpodivin:
The `PromptManager` is now primary handler of all prompts. The constants that used to be defaults were removed, as well as the old prompts defined in `prompts.yml`.  The new setup is less forgiving, if templates are missing for some reason, an exception will be raised, rather than defaults loaded.

- To facilitate testing, the `.env` file was renamed to `env_file`. The change is reflected in `docker-compose.yaml` and docs.
- `logdetective/utils.py` was formatted with ruff
- `PromptConfig` now only holds optional information for building prompts, we may have to do something with it in the future.


logdetective/models.py
```diff
@@ -2,21 +2,10 @@
...
 
 class PromptConfig(BaseModel):
     """Configuration for basic log detective prompts."""
 
-    prompt_template: str = PROMPT_TEMPLATE
-    agent_start_prompt: str = AGENT_START_PROMPT
-
-    default_system_prompt: str = DEFAULT_SYSTEM_PROMPT
-
     references: Optional[list[dict[str, str]]] = None
```
Jany26, line 9: Does it make sense keeping `PromptConfig` around when it only has one attribute which is still loaded into `PromptManager._references`?

logdetective/prompts.yml
Jany26, line 11: Since this file no longer contains prompts of any kind and serves only to store `references`, I think the naming or placement might be misleading... I would put this into `logdetective/prompts/` directory, but that would probably be backwards-incompatible 🤔

Jany26 (approved):
I only have 2 things to consider - not functional, mostly cosmetic. Otherwise looks good.

logdetective/models.py
jpodivin: Yeah it's almost redundant. We may eventually merge it into general config, if we ever get rid of CLI, or expand it. But I would still keep it around for now.

logdetective/prompts.yml
jpodivin: The `logdetective/prompts` is really for templates though, not for config files. It mostly comes down to personal preference.

PR #527, opened 2026-05-21
Make extractor SkipSnippets None by default

Jany26:
Improves upon https://github.com/fedora-copr/logdetective/pull/525.

### DIFF
--- a/logdetective/utils.py
+++ b/logdetective/utils.py
@@ -285,9 +285,14 @@ def prompt_to_messages(
     return messages
 
 
-def filter_snippet_patterns(snippet: str, skip_snippets: SkipSnippets) -> bool:
+def filter_snippet_patterns(
+    snippet: str,
+    skip_snippets: Optional[SkipSnippets] = None
+) -> bool:
     """Try to match snippet agains provided patterns to determine if we should
     filter it out or not."""
+    if not skip_snippets:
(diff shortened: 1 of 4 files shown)


logdetective/utils.py
```diff
@@ -285,9 +285,11 @@ def prompt_to_messages(
     return messages
 
 
-def filter_snippet_patterns(snippet: str, skip_snippets: SkipSnippets) -> bool:
+def filter_snippet_patterns(snippet: str, skip_snippets: Optional[SkipSnippets]) -> bool:
```
jpodivin: `skip_snippets` should have `None` as default, if it is optional.

logdetective/utils.py
```diff
@@ -296,7 +298,7 @@ def filter_snippet_patterns(snippet: str, skip_snippets: SkipSnippets) -> bool:
     return False
 
 
-def load_skip_snippet_patterns(path: str | None) -> SkipSnippets:
+def load_skip_snippet_patterns(path: str | None) -> Optional[SkipSnippets]:
```
jpodivin: I would prefere union with `None` over optional here. It is more precise.

jpodivin (changes requested):

jpodivin (approved):

PR #528, opened 2026-06-01
Try getting logs without size headers

Jany26:
This change affects only RemoteLogs - requests that contain URLs. Putting log content directly into the request still requires Content-Length header.
- Content-Length is present and too large -> `RemoteLogTooLargeError`
- Content-Length is present and invalid -> `RemoteLogHeaderError`
- Content Length is present and okay -> stream the file and continually check size (in case header lies about size)
- Content Length is not-present -> stream the file and continually check size (this fixes what OBS guys suggested)

### DIFF
--- a/logdetective/remote_log.py
+++ b/logdetective/remote_log.py
@@ -66,7 +66,9 @@ class RemoteLog:
         return True
 
     async def get_url_content(self) -> str:
-        """Validate log url, check the content size from header, and return log text."""
+        """Validate log url, check content size (either using Content-Length, or,
+        if missing, during file reading), and return log text.
+        """
         if not self.validate_url():
             LOG.error("Invalid URL received ")
             raise RemoteLogRequestError(f"Invalid log URL: {self.url}")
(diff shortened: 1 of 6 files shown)


logdetective/remote_log.py
```diff
@@ -82,22 +82,38 @@ async def get_url_content(self) -> str:
...
+                "No Content-Length header for %s; enforcing size limit while reading", self.url
             )
-        self.remote_log_size = size_check.size_in_bytes
-        # if size-check passes, we obtain the whole content
         try:
             response = await self._http_session.get(self.url, raise_for_status=True)
         except (aiohttp.ClientResponseError, aiohttp.ClientConnectorError) as ex:
             raise RemoteLogAccessError(f"We couldn't obtain the log from {self.url}") from ex
-        return await response.text()
+        return await self._read_with_size_limit(response)
```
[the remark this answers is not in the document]
jpodivin: This is a valid concern.

jpodivin (changes requested):
Valid issue was identified by the automated review.

logdetective/server/gitlab.py
jpodivin, line 318: Please adjust the docstring to reflect the new conditional on line 349

jpodivin (changes requested):
Are original fields of the `ContentSizeCheck` still used for anything? In any event, we should change it, so that it reflects the new usage.

logdetective/server/gitlab.py
Jany26: ad ContentSizeCheck fields: yes, both fields are used: for logging and for selecting the proper RemoteLog/HTTP exception

logdetective/utils.py
```diff
@@ -348,25 +344,32 @@ class ContentSizeCheck(NamedTuple):
...
 
-    result: bool
-    value_present: bool
+    proceed: bool
     size_in_bytes: int | None
 
 
-def check_content_size(headers: dict[str, str], size_limit: int) -> ContentSizeCheck:
+def check_content_size(
+    headers: dict[str, str],
```
jpodivin: It's not that much of a problem. But it would be better to use the actual type `Headers` from starlette, for annotation, or `Mapping`, because the argument is never really a header, and there are important differences.

jpodivin (changes requested):
One last nitpick and I think we are good to go.

jpodivin (approved):

PR #535, opened 2026-06-08
Provide minimal information about build environment

jpodivin:
Requests coming from gitlab now have commentary attached, explaining usage of Mock and Koji.
The file was also reformatted using ruff for consistency.

### DIFF
--- a/logdetective/server/gitlab.py
+++ b/logdetective/server/gitlab.py
@@ -30,6 +30,7 @@ from logdetective.server.models import (
     GitLabInstanceConfig,
     JobHook,
     APIResponse,
+    BuildMetadata,
 )
 from logdetective.server.database.models import (
     AnalyzeRequestMetrics,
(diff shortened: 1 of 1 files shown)


logdetective/server/gitlab.py
```diff
@@ -337,8 +351,7 @@ async def check_artifacts_file_size(
         ) from ex
     LOG.debug("Checking content-length for %s", artifacts_path)
     size_check: ContentSizeCheck = check_content_size(
-        head_response.headers,
-        gitlab_cfg.max_artifact_size
+        head_response.headers, gitlab_cfg.max_artifact_size
```
Jany26, line 354: I don't get why ruff split the arguments in other calls to multiple lines, but merged arguments in this call to the same line :D Anyway, LGTM.

Jany26 (approved):

logdetective/server/gitlab.py
jpodivin: I don't get it either :D

PR #537, opened 2026-06-10
Restore models import

jpodivin:
Without import of the models module, the revision generation created changes that make no sense, such as removal of existing tables etc.

This is because without the import, models are not executed, alembic doesn't know about them and then tries to reconcile the database state with the only model it has, which is `Base`.


majamassarini (approved):
LGTM thanks for the fix!

PR #544, opened 2026-06-25
Set version to 4.8.0

jpodivin:
We have more than enough changes to release a new version.


Jany26 (approved):
It makes total sense to do a new release with the caching and vertexai support.

PR #546, opened 2026-06-25
Enable client side caching of tool calls and LLM responses

jpodivin:
With this modification we can cache both tool calls and LLM responses without having to change anything about our inference backend. The cache for LLM calls is set to hold 35 calls by default, this should be enough for most runs, while not excessively taxing memory. 

The tool calls cache is not constrained at all, since number of tool calls is extremely limited by constraints.

### DIFF
--- a/logdetective/server/models.py
+++ b/logdetective/server/models.py
@@ -229,6 +229,8 @@ class InferenceConfig(BaseModel):  # pylint: disable=too-many-instance-attribute
     # Provider-specific settings passed to the beeai backend via the settings dict.
     # Keys are beeai internal setting names (e.g. api_key, base_url, vertex_project, ...).
     provider_settings: dict[str, str] = {}
+    # Harness cache settings
+    llm_call_cache_size: int = 35
 
 
 class ExtractorConfig(BaseModel):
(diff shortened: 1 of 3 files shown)


logdetective/server/models.py
```diff
@@ -229,6 +229,8 @@ class InferenceConfig(BaseModel):  # pylint: disable=too-many-instance-attribute
     # Provider-specific settings passed to the beeai backend via the settings dict.
     # Keys are beeai internal setting names (e.g. api_key, base_url, vertex_project, ...).
     provider_settings: dict[str, str] = {}
+    # Harness cache settings
+    llm_call_cache_size: int = 35
```
Jany26, line 233: What does the number 35 stand for?

Jany26 (approved):
Just missing an explanation for the config value, other than that LGTM.

logdetective/server/models.py
jpodivin: It's the cache size.

Jany26: Nevermind, apparently it is the individual LLM prompt-output, or tool call-result. Sorry for not reading the title of the PR properly :smile:. But still, it might help avoid confusion in the future (as I've just brilliantly demonstrated 😅) to also somehow mention this in the code comment.

PR #549, opened 2026-07-07
Revert "Add LLM call cache configuration"

jpodivin:
The sliding cache introduced an unexpected issue of delimiter leak inside of tool calls. For example:

```
   - Error: 1 validation error for AvailableTools
     item
     Field required [type=missing, input_value={'artifact_name': '|>builder-live.log.gz<|'}, input_type=dict]
     For further information visit https://errors.pydantic.dev/2.13/v/missing
```


Jany26 (approved):
too bad beeai is broken in the most unexpected ways...

PR #550, opened 2026-07-08
Fix typo in the var name

jpodivin:


Jany26 (approved):
for some weird reason, djlint is panicking...

PR #553, opened 2026-07-13
Add config flag for solution presence

Jany26:
Solution is an experimental feature and it might not be desirable to have it present in the response, so we want a fast way of disabling its presence in the response.
The dependant services are:
- packit-dashboard (which are also linked log detective jobs from koji CI),
- and possibly logdetective website (although the website does not access solution or show it).

This behavior is affected by the `generate_solution` flag inside the general config in `config.yml`.
- default behavior is "true", i.e. show solution (string),
- overriding with "false" means the field is None.

It should be sufficient to just change the flag and restart the server container.
Tested on a local compose.

## Summary by CodeRabbit

* **New Features**
  * Added a `generate_solution` configuration option to control whether solution content is included in API responses.
  * Solutions are included by default, and are suppressed when the option is disabled.
* **Documentation**
  * Updated configuration guidance describing the new `generate_solution` setting (default: enabled).
* **Tests**
  * Added tests confirming the response omits the solution when disabled and includes it when enabled.

### DIFF
--- a/logdetective/server/agent/agent.py
+++ b/logdetective/server/agent/agent.py
@@ -191,4 +191,6 @@ async def analyze_artifacts(
         no_issue_found=structured_output.no_issue_found,
         snippets=all_snippets,
     )
+    if not SERVER_CONFIG.general.generate_solution:
+        response.solution = None
     return response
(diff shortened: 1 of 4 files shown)


jpodivin (approved):
After the typo is fixed we can merge this.

jpodivin (approved):

PR #556, opened 2026-07-14
Fix type hints of the method

jpodivin:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the reported return type for traceback truncation, improving consistency and reliability for integrations that inspect type information.


Jany26 (dismissed):

Jany26 (commented):
On second thought, I think it's actually fine as it is, it transforms one snippet, and then is used in a map().

Jany26 (approved):
your change is actually correct :D sorry for the confusion

PR #560, opened 2026-07-17
Populate the annotations DB tables with data collected from website

Jany26:
The website now has the `/download?since=YYYY-MM-DD` functionality, so we are ready to start testing the vector db update script.

It required adding a new DB table for updates (at least ones which add some contributions) - one update stores info about:
- the date of running the script (so that next time, we know what do put into `since=YYYY-MM-DD`)
- how many json files were actually processed (it is possible that we process non-zero contribution json files with annotations and not put any new entries, as we don't use full timestamps but only dates, but the frequency of contributions has mostly died down, so I'm not expecting any such scenario).

There is an additional guard for putting stuff into the database, and it is the `source_path` attribute of the `AnnotatedBuilds` table. While we go through each json, we check if there is any build with the same path as the relative path from the archive to the annotation json file.

There is also `--dry-run` (does not affect database) and `--reset` (wipe all annotation-related tables and start over with populating the DB).

While it is possible to run the script by itself, it is supposed to be run as a cronjob (monthly should be sufficient) either as an openshift pod or as a container inside a compose.

Archive content validation uses pydantic models for simplicity.

Some notes:
- On my T14, the script takes a little under 10 minutes.
- As of 2026-07-21, we have 1170 valid annotated builds and 2259 valid annotated snippets which pass validation and would be added to the db.
- Tested on live local compose, and using `AnnotatedSnippets.get_by_snippet_embedding` returns snippets inline with expectations.

## Summary by CodeRabbit

* **New Features**
  * Added automated synchronization of user-contributed build annotations from downloadable archives.
  * Added incremental updates, initial bootstrap, duplicate detection, reset, and dry-run options.
  * Added validation for contribution data and improved handling of malformed archive entries.
  * Added tracking for processed files, update dates, annotated builds, and snippets.
  * Added deployment configuration for scheduled contribution updates.
* **Tests**
  * Added coverage for validation, synchronization modes, deduplication, reset behavior, and semantic lookup.


logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,472 @@
...
+
+    logger.info("--- Current update ---")
+    logger.info("JSONs processed: %d", current.jsons_processed)
+    logger.info("Builds added: %d", current.builds_added)
+    logger.info("Snippets added: %d", current.snippets_added)
+    logger.info("--- Overall DB state ---")
+    logger.info("Total annotated builds: %d", build_count)
+    logger.info("Total files processed across all updates: %d", total_processed)
+    if duplicates_skipped > 0:
+        logger.info("Duplicate entries skipped (cumulative): %d", duplicates_skipped)
```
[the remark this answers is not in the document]
Jany26: there's currently no entries in the DB under AnnotatedBuilds, so no legacy builds with source_path=null

files/openshift-sync-cronjob.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: update-db-web-contributions
+spec:
+  schedule: "15 2 * * 1"  # Every Monday at 2:15 AM
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 1
+  failedJobsHistoryLimit: 3
+  jobTemplate:
+    spec:
+      backoffLimit: 2
+      activeDeadlineSeconds: 1800  # TODO: full reset may need more than 30 minutes
```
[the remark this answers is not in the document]
Jany26: need to test this to see how much time we'll require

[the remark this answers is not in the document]
Jany26: this is just a first guess

logdetective/server/database/update_db_web_contributions.py
```diff
@@ -0,0 +1,472 @@
...
+    async with transaction(commit=True) as session:
+        await session.execute(delete(AnnotatedSnippets))
+        await session.execute(delete(AnnotatedBuilds))
+        await session.execute(delete(AnnotationUpdates))
+    logger.info("All annotation tables wiped.")
+
+
+async def validate_and_ingest(
+    contributions: dict[str, dict],
+    url: str,
```
jpodivin: Why isn't URL used?

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+async def get_existing_source_paths() -> set[str]:
+    """Query all existing source_path values from AnnotatedBuilds."""
+    async with transaction(commit=False) as session:
+        result = await session.execute(
+            select(AnnotatedBuilds.source_path)
+            .where(AnnotatedBuilds.source_path.isnot(None))
+        )
+        return set(result.scalars().all())
```
jpodivin: The `all` method can also return `None`. Suppose that all rows have `source_path == None`

logdetective/server/web_contributions_update.py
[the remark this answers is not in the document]
Jany26: functionally this will stay, but i will adjust the comments and docs accordingly

docker-compose.yaml
```diff
@@ -70,6 +70,11 @@ services:
     ports:
       - "${POSTGRESQL_PORT:-5432}:${POSTGRESQL_PORT:-5432}"
     env_file: ./env_file
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRESQL_USER:-postgres}"]
+      interval: 60s
+      timeout: 10s
+      retries: 5
```
jpodivin: Dependency on this should be added to the primary server container as well.

logdetective/server/web_contributions_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    contributions: dict[str, dict] = {}
+    with tarfile.open(fileobj=archive_file, mode="r:gz") as tar:
+        for member in tar.getmembers():
+            if not member.isfile() or not member.name.endswith(".json"):
+                continue
+            if not _is_safe_tar_member(member.name):
+                logger.warning("Skipping unsafe tar member path: %s", member.name)
+                continue
+            if not member.name.startswith("results/results/"):
+                logger.debug("Skipping %s: not under results/results/", member.name)
```
jpodivin: Info level would be better.

logdetective/server/user_contributions_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+            if not _is_safe_tar_member(member.name):
+                logger.warning("Skipping unsafe tar member path: %s", member.name)
+                continue
+            if not member.name.startswith("results/results/"):
+                logger.debug("Skipping %s: not under results/results/", member.name)
+                continue
+            try:
+                f = tar.extractfile(member)
+                if f is None:
+                    continue
```
jpodivin, line 221: Log here as well, empty result is irregular.

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,439 @@
...
+    parse_archive,
+    deduplicate,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _sanitize_for_pg(value: str) -> str:
+    """Strip null bytes — the only byte PostgreSQL rejects in UTF-8 text columns."""
+    return value.replace("\x00", "")
```
jpodivin: Does this happen often?

logdetective/server/web_contributions_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+    if not match:
+        raise RuntimeError(
+            f"Could not extract date from archive filename: {filename!r}"
+        )
+    return date.fromisoformat(match.group("archive_date"))
+
+
+def _is_safe_tar_member(name: str) -> bool:
+    """Reject tar member paths that could escape the archive root."""
+    return not (name.startswith("/") or ".." in name)
```
jpodivin: Split the path first and check on parts.

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,439 @@
...
+            snippets_added=self.snippets_added + other.snippets_added,
+        )
+
+
+DOWNLOAD_URL = "https://logdetective.com/download"
+MAX_DOWNLOAD_BYTES = 500 * 1024 * 1024
+DOWNLOAD_TIMEOUT = aiohttp.ClientTimeout(total=600)
+
+
+async def _download_once(
```
jpodivin: We should retry here if we get common connection errors.

logdetective/server/web_contributions_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+
+    if not isinstance(data["logs"], dict):
+        logger.warning("Skipping %s: 'logs' is not a dict", file_path)
+        return None
+
+    all_snippets: list[SnippetData] = []
+    for log_name, log_entry in data["logs"].items():
+        extracted = _extract_snippets(log_entry, log_name, file_path)
+        if extracted is None:
+            return None
```
jpodivin: We probably shouldn't throw all logs away if one is malformed.

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,439 @@
...
+        "Inserted %d builds with %d snippets",
+        len(valid), snippet_count,
+    )
+
+    return CurrentUpdateStats(archive_json_count, len(valid), snippet_count)
+
+
+async def run_update(url: str, dry_run: bool = False, reset: bool = False) -> None:
+    """Main update orchestrator."""
+    from logdetective.server.database.base import get_pg_url  # pylint: disable=import-outside-toplevel
```
jpodivin: Why not import it on the top level?

logdetective/server/web_contributions_helpers.py
```diff
@@ -0,0 +1,165 @@
...
+def deduplicate(
+    contributions: list[ContributionData],
+    existing_paths: set[str],
+) -> list[ContributionData]:
+    """Filter out contributions whose source_path is already in the DB."""
+    new = [c for c in contributions if c.source_path not in existing_paths]
+    skipped = len(contributions) - len(new)
+    if skipped:
+        logger.info("Skipped %d already-imported contributions", skipped)
+    return new
```
jpodivin: This could be an sql query.

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,439 @@
...
+
+            disposition = resp.headers.get("Content-Disposition", "")
+            logger.info("Content-Disposition: %s", disposition or "(not set)")
+            filename = (
+                disposition.split("filename=")[-1].strip('"')
+                if "filename=" in disposition
+                else ""
+            )
+
+            spool = tempfile.SpooledTemporaryFile(  # pylint: disable=consider-using-with
```
jpodivin: This should really be a context manager instead. Especially if other exceptions show up.

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,439 @@
...
+            return paths
+    except Exception as exc:
+        logger.error("Failed to query existing source paths: %s", exc)
+        raise
+
+
+@backoff.on_exception(backoff.expo, OperationalError, max_tries=DB_MAX_RETRIES)
+async def insert_contributions(
+    contributions: list[ContributionData],
+    embeddings: list[np.ndarray],
```
jpodivin, line 87: Each snippet should have an embedding already associated, in the datastructure, maybe as a tuple (snippet, embedding), that way we won't need to keep track of `embed_idx`.

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,439 @@
...
+    return embed_idx
+
+
+async def reset_tables() -> None:
+    """Wipe all rows from AnnotatedSnippets, AnnotatedBuilds, and AnnotationUpdates."""
+    async with transaction(commit=True) as session:
+        await session.execute(delete(AnnotatedSnippets))
+        await session.execute(delete(AnnotatedBuilds))
+        await session.execute(delete(AnnotationUpdates))
+    logger.info("All annotation tables wiped.")
```
jpodivin: Raise to warning please, this is a radical action.

jpodivin (changes requested):
Tests and linters are failing, please take a look at them.

logdetective/server/web_contributions_update.py
Jany26: sorry, that was a leftover from debugging

logdetective/server/web_contributions_update.py
Jany26:
```bash
$ find . -type f -name "*.json" -exec grep -l '\\u0000' {} +
./results/2024-09-28/copr/8088229/1727531619.json
./results/2025-06-11/copr/9148659/46d34799-0782-47bd-8ff5-3d73e8b34194.json
./results/2025-06-11/copr/9148659/87dbf389-4181-4b1e-a753-1f849fd9c2d4.json
./results/2025-06-16/copr/9148659/34789ce8-29cb-41f8-9dd2-09d63e8797a6.json
./results/2025-06-16/copr/9148659/ac0f294d-d382-45d9-854a-0da7ab74e070.json
./reviews/parsed/1727531619-1738930172.json
./reviews/parsed/46d34799-0782-47bd-8ff5-3d73e8b34194-1749641804.json
```

I guess this should have been done as part of the sanitization...

docker-compose.yaml
Jany26: after discussion, healthcheck is out-of-scope for this PR, so we will do the wait using the established await_psql script

logdetective/server/web_contributions_helpers.py
Jany26: good idea, apparently using `sqlalchemy.dialects.postgresql.pg_insert` is the best for this (it can handle `ON CONFLICT DO NOTHING`). this requires different preprocessing (creating map `source_path` -> list of (snippet, embedding))

Jany26:
@coderabbitai review

logdetective/server/web_contributions_helpers.py
```diff
@@ -0,0 +1,156 @@
...
+        ))
+    return snippets
+
+
+def validate_contribution(data: dict, file_path: str) -> ContributionData | None:
+    """
+    Validate a contribution JSON and extract all snippets.
+
+    Returns a ContributionData if valid, None otherwise.
+    """
```
jpodivin: The above comment by rabbit is correct, and it strikes me that we should leverage Pydantic validation here.

jpodivin (changes requested):

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,464 @@
...
+                        "Download exceeded size limit (%d bytes), aborting",
+                        MAX_DOWNLOAD_BYTES,
+                    )
+                    raise RuntimeError(
+                        f"Archive download exceeded {MAX_DOWNLOAD_BYTES} bytes, aborting."
+                    )
+                spool.write(chunk)
+            spool.seek(0)
+            logger.info("Downloaded %d bytes (%s)", total, filename)
+            return filename
```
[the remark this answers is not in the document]
jpodivin: This is also something that needs a look.

logdetective/server/web_contributions_helpers.py
```diff
@@ -0,0 +1,156 @@
...
+            )
+            return None
+        if not snippet["text"] or not snippet["user_comment"]:
+            continue
+        snippets.append(SnippetData(
+            text=snippet["text"],
+            user_comment=snippet["user_comment"],
+            log_name=log_name,
+        ))
+    return snippets
```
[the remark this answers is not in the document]
Jany26: we'll rework the whole validation to use pydantic models

logdetective/server/web_contributions_helpers.py
Jany26: yeah using pydantic models really simplifies things, this function will be redundant

logdetective/server/web_contributions_helpers.py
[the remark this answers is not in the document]
Jany26: this whole thing will be rewritten to use pydantic

Jany26:
@coderabbitai review

logdetective/server/web_contributions_update.py
```diff
@@ -0,0 +1,323 @@
...
+            url, since=archive_date_from_filename,
+        )
+        await validate_and_insert_contributions(
+            delta_contributions, embedding_model, stats, dry_run,
+        )
+    else:
+        # Incremental: just fetch since last update
+        logger.info("Incremental sync since %s.", latest_date)
+        # contributions, _ = await fetch_and_parse(url, since=latest_date)
+        contributions, _ = await fetch_and_parse(url)
```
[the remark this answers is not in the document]
Jany26: alright alright i forgot to remove some debugging stuff

jpodivin (approved):

Jany26:
Made some final minor simplifications and notably made `AnnotationBuilds.source_path` mandatory.

This is ready to merge, but there seems to be some transient github action error with `run_liters` and `run_pytest`. Checks are passing locally and e2e testing on a live-compose also works.

EDIT:
- even after 5 days the github action error persists, might be worth looking into -> rebasing on https://github.com/fedora-copr/logdetective/pull/564 should fix this
- whole when ran in a container, got OOM killed -> so I rewrote it to use pipelines and do embeddings and db inserts in batches and reduced `6GiB` RAM usage to around `<1.2GiB`, while not sacrificing runtime significantly
- now the script on whole archive from end of June (1170 builds, 2259 snippets) runs 15-17mins

Jany26:
@coderabbitai review

docker-compose.yaml
```diff
@@ -85,6 +85,30 @@ services:
...
+    env_file: ./env_file
+    working_dir: /src
+    volumes:
+      - .:/src/:z
+    entrypoint:
+      - bash
+      - -c
+      - "python -m logdetective.server.web_contributions_update"
+    profiles:
+      - sync
```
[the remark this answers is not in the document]
Jany26: Doing `pg_isready` checks here renders `await_psql` script useless I guess. And even then, this container will not be started by itself but in a live-compose, so checking if those containers have started should be enough to proceed. If database is not ready, the script will fail at the start anyway.

alembic/versions/b266cb6cf7d5_add_annotation_updates_and_source_path_.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    op.add_column(
+        'annotated_builds',
+        sa.Column(
+            'source_path',
+            sa.String(),
+            nullable=False,
+            comment='Path to the contribution in the archive'
+        ),
+    )
```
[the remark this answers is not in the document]
Jany26: No builds were added into the database between that migration and now.

logdetective/server/database/models/annotated_builds.py
```diff
@@ -109,15 +124,80 @@ class AnnotatedBuilds(Base):
...
+    async def get_latest_archive_date(cls) -> Optional[date]:
+        """Return the archive_date of the most recent revision, or None."""
+        try:
+            async with transaction(commit=False) as session:
+                result = await session.execute(
+                    select(cls.archive_date).order_by(cls.id.desc()).limit(1)
+                )
+                return result.scalar_one_or_none()
+        except ProgrammingError:
+            return None
```
jpodivin: This should be logged at least. And it must be commented, explaining why are we not raising `ProgrammingErrors`.

jpodivin (changes requested):

logdetective/server/database/models/annotated_builds.py
Jany26: Good catch!

I think this check is probably redundant. I probably added this during live testing when the update table was actually missing. This can happen when update script runs before migrating properly/fully (or running the script on a stale container).

It may be better to remove the try-except and actually traceback in this case!

Jany26: Removed redundant try-except.

logdetective/server/models.py
```diff
@@ -479,3 +479,46 @@ class MetricResponse(BaseModel):
...
+        """Keep only well-formed snippet dicts; silently drop the rest."""
+        if not isinstance(v, list):
+            return []
+        return [
+            i for i in v if isinstance(i, dict) and i.get("text") and i.get("user_comment")
+        ]
+
+
+class ContributionBuild(BaseModel):
+    """Full web contribution (JSON) with annotated build failure and snippets."""
```
jpodivin: Please modify comments so that it's clear where the contributions come from.

docker-compose.yaml
```diff
@@ -85,6 +85,27 @@ services:
...
 
+  web-contributions-update:
+    image: quay.io/logdetective/server:latest
+    build:
+      context: .
+      dockerfile: ./Containerfile
+    env_file: ./env_file
+    working_dir: /src
+    volumes:
+      - .:/src/:z
```
jpodivin: Replace with log mount please.

jpodivin (changes requested):

jpodivin (approved):

docker-compose.yaml
Jany26: i chose `./log/user_contribution_update.log` as the local path for update script logs, similar to server logs (`./log/logdetective.log`)

Jany26:
@coderabbitai review

Jany26:
Merging now, before we actually run it in deployment though, we'll have to fix our issues with data persistence on volume.

Jany26:
There is a potential improvement in how we handle embedding long snippets, either we:
- split the long snippet into overlapping chunks, embedding each one separately and linking a list of embeddings to one snippet (instead of just 1), or
- embed just a fraction of the original snippet with only critical information extracted (perhaps by some LLM).
Used embedding model has a max length of 512 tokens (~2000 characters). Some snippets reach lengths of tens of thousands (especially if a user selected a whole overly-verbose test report for example).

PR #562, opened 2026-07-21
Truncate, rather than split, snippets.

jpodivin:
We used to split snippets that were over given length into multiple sub-snippets. But this posed certain issues when working with SnippetAnalysisTool. Multiple snippets technically occupied the same line and could be analyzed only once.
At the same time, the agent could get confused.

Instead, we are going to truncate, much like in the python traceback extractor. This does mean that in certain limit cases, when the length of the truncation marker is greater than the maximum snippet length, we would effectively replace snippets entirely with truncation. To prevent that, a constraint will be set on the parameter in the config model, and raise an exception if utility function is called with an unreasonable argument value.

I've also expanded the tests, so that we verify the chunk generator behavior.

## Summary by CodeRabbit

* **New Features**
  * Long log snippets are now clearly marked with a `<truncated>` tag.
  * Extractor tool guidance now explains how truncated snippets are identified.
* **Improvements**
  * Snippet length settings now include clearer descriptions and enforce a minimum length.
  * Oversized snippets are truncated consistently instead of being split repeatedly.
* **Bug Fixes**
  * Invalid snippet length configurations now return a clear validation error.

### DIFF
--- a/logdetective/constants.py
+++ b/logdetective/constants.py
@@ -29,3 +29,9 @@ PROMPT_PATH = os.environ.get(
 
 # Vector size of BAAI/bge-small-en-v1.5
 EMBEDDING_VECTOR_SIZE = 384
+
+# Marker for snippet truncation
+TRUNCATED = "<truncated>"
+
+# Minimum length of a snippet
+MINIMUM_SNIPPET_TRUNCATION_LEN = 100
--- a/logdetective/server/agent/tools.py
+++ b/logdetective/server/agent/tools.py
@@ -152,6 +152,7 @@ class DrainExtractorTool(ExtractorTool):
         "Use this tool at most once per artifact."
         "Extracts up to {max_clusters} snippets from a log file, using clustering Drain algorithm."
         "Maximum length of extracted snippet is {max_snippet_len}."
+        "Truncated snippets are marked with the `<truncated>` tag."
     )
     extractor: DrainExtractor
 
@@ -188,6 +189,7 @@ class CSGrepExtractorTool(ExtractorTool):
         "Extracts up to {max_clusters} snippets from a log file containing GCC traces, using csgrep tool."
         "Do not use on artifacts that don't contain traces from compiler."
         "Maximum length of extracted snippet is {max_snippet_len}."
+        "Truncated snippets are marked with the `<truncated>` tag."
     )
 
     extractor: CSGrepExtractor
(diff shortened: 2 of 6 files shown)


logdetective/server/agent/tools.py
```diff
@@ -152,6 +152,7 @@ class DrainExtractorTool(ExtractorTool):
         "Use this tool at most once per artifact."
         "Extracts up to {max_clusters} snippets from a log file, using clustering Drain algorithm."
         "Maximum length of extracted snippet is {max_snippet_len}."
+        "Truncated snippets are marked with the `<truncated>` tag."
```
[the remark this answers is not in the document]
jpodivin: This is a preestablished pattern. We can change it later.

logdetective/constants.py
```diff
@@ -29,3 +29,9 @@
 
 # Vector size of BAAI/bge-small-en-v1.5
 EMBEDDING_VECTOR_SIZE = 384
+
+# Marker for snippet truncation
+TRUNCATED = "<truncated>"
+
+# Minimum length of a snippet
+DEFAULT_MINIMUM_SNIPPET_LEN = 100
```
Jany26: The var name seems confusing, since it is not really a minimum snippet length, but rather the "minimum truncation limit/marker length" or something along those lines.

logdetective/utils.py
```diff
@@ -108,18 +110,20 @@ def get_chunks(
...
-                chunk = remainder
-
+            # If the chunk is too long, truncate and add <truncated> tag
+            chunk = chunk[:max(max_chunk_len - len(TRUNCATED), 0)] + TRUNCATED
+            yield (original_line, chunk)
+            chunk = ""
     # if we still have some text left over
-    yield (original_line, chunk)
+    if chunk:
+        yield (original_line, chunk)
```
Jany26, line 126: Won't this create snippets with the same line number still?

Jany26 (changes requested):
Not sure about a couple of things, otherwise looks good.

logdetective/constants.py
jpodivin: True

logdetective/utils.py
jpodivin: I don't think so. This statement can only be reached if there is a some text in `chunk`, and we have finished iterating over entire list of lines.

This means that we didn't hit conditional on line `119`, which sets `chunk = ""`, nor have we hit any of the statements changing `original_line`. It would have to be continuation of an old message, `106`, but `chunk` would have to already be filled with some text, lines `117` to `118`. At the same time, it would have to be under `max_chunk_len`, as that would reset the chunk to empty, line `123`. It would also have to be a last line in the `lines` list.

But in this situation, we would not hit any `yield` since setting `original_line`. Therefore, no duplicates.

Jany26: okay cool!

logdetective/constants.py
Jany26: thanks!

Jany26 (approved):

PR #568, opened 2026-08-03
Annotation DB lookup tool for the agent

Jany26:
We made the script for populating the database with user contributions from website. We fixed the database persistence issues. Now it is time to actually provide the data to the agent.

DB lookup tool is controlled using 2 general options in `server/config.yml`: 
- `annotation_lookup_tool` - flag for enabling the tool, false by default
- `top_k_annotations` - how many best-match snippets are returned by the DB (default: 2, more would probably be wasteful, needs testing and fine-tuning)

I still need to thoroughly live-test (maybe adjust the tool description) the tool before merging, but we can already start the review process.

## Summary by CodeRabbit

* **New Features**
  * Added optional annotated-snippet lookup using semantic search.
  * Search results can include matching text, annotations, source artifacts, build problems, and solutions.
  * Added configuration options to enable lookup and control the number of results returned.
* **Bug Fixes**
  * Added handling for empty searches and reporting of embedding or lookup failures.
* **Tests**
  * Added coverage for successful searches, empty results, configuration behavior, and error scenarios.

### DIFF
--- a/logdetective/server/agent/agent.py
+++ b/logdetective/server/agent/agent.py
@@ -21,14 +21,21 @@ from litellm.exceptions import (
 from pydantic import ValidationError
 
 from logdetective.remote_log import RemoteLog
-from logdetective.server.config import PROMPT_CONFIG, SERVER_CONFIG, SKIP_SNIPPETS_CONFIG
+from logdetective.server.config import (
+    PROMPT_CONFIG,
+    SERVER_CONFIG,
+    SKIP_SNIPPETS_CONFIG,
+    EMBEDDING_MODEL_INSTANCE,
+)
 from logdetective.server.agent.tools import (
     ExtractorTool,
     DrainExtractorTool,
     CSGrepExtractorTool,
     PythonTracebackExtractorTool,
     SnippetAnalysisTool,
+    AnnotatedSnippetLookupTool,
 )
+from logdetective.server.database.models.annotated_builds import AnnotatedSnippets
 from logdetective.server.models import APIResponse, BuildMetadata, AgentResponse
 from logdetective.server.exceptions import (
     LogDetectiveAgentResponseFailure,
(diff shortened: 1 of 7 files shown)


server/config.yml
```diff
@@ -91,3 +91,7 @@ general:
   # Enable lazy fetching of build artifacts
   # delay_artifact_download: true
   agent_timeout: 600  # Timeout on agent execution
+  # Enable DB lookup of annotated snippets (requires populated annotated_snippets table)
+  # annotation_lookup_tool: true
+  # How many top annotated snippet searches to return
+  # top_k_annotations: 2
```
jpodivin: This should be an upper constraint on a parameter set by agent.

logdetective/server/agent/tools.py
```diff
@@ -349,3 +351,103 @@ def _create_emitter(self) -> Emitter:
     @property
     def input_schema(self) -> type[SnippetAnalysisToolInput]:
         return SnippetAnalysisToolInput
+
+
+class AnnotatedSnippetLookupToolInput(BaseModel):
```
jpodivin, line 357: We should let agent ask for a specific number of results. Subject to reasonable constraints.

logdetective/server/agent/agent.py
```diff
@@ -122,6 +129,22 @@ async def analyze_artifacts(
             )
         )
 
+    if EMBEDDING_MODEL_INSTANCE and await AnnotatedSnippets.get_count() > 0:
+        tools.append(
+            AnnotatedSnippetLookupTool(
+                EMBEDDING_MODEL_INSTANCE,
```
jpodivin: Is there any situation when the tool could be initialized with any other embedding model? If not, we shouldn't use it as an arg.

logdetective/server/agent/agent.py
```diff
@@ -122,6 +129,22 @@ async def analyze_artifacts(
             )
         )
 
+    if EMBEDDING_MODEL_INSTANCE and await AnnotatedSnippets.get_count() > 0:
+        tools.append(
+            AnnotatedSnippetLookupTool(
```
jpodivin: Please add cache, either here, or elsewhere in the call stack, so we don't use the model more than necessary.

jpodivin (changes requested):

logdetective/server/agent/tools.py
Jany26: Oh okay, I get it now. I'll add another field to the tool input, restricted to be 1 <= x <= `SERVER_CONFIG.general.max_annotations` (renaming top_k_annotations option to max_annotations).

server/config.yml
Jany26: So if I understand correctly, we want something like `max_annotations` and coupled with your another review point, the agent will decide the number of lookup results in case lookup tool is called.

logdetective/server/agent/agent.py
Jany26: Right, an oversight.

jpodivin (approved):
I like this.

PR #584, opened 2026-08-19
Limit thread usage by embedding model

Jany26:
`threads` controls the relevant `{intra,inter}_op_num_threads` config
https://onnxruntime.ai/docs/performance/tune-performance/threading.html
https://github.com/qdrant/fastembed/blob/main/fastembed/common/onnx_model.py#L106-L108

With this change, there will be no way to get the default (None) behavior which starts onnx runtime on all available CPU cores (if no GPU can be found).

## Summary by CodeRabbit

* **New Features**
  * Added a configurable setting for controlling the number of threads used by the embedding model.
  * The setting defaults to 4 threads and accepts only positive whole numbers.

* **Improvements**
  * Embedding model processing now consistently honors the configured thread count, helping administrators tune resource usage and performance.
  * Configuration documentation now explains thread usage and model loading behavior.

### DIFF
--- a/logdetective/server/user_contributions_update.py
+++ b/logdetective/server/user_contributions_update.py
@@ -74,6 +74,7 @@ DOWNLOAD_URL = "https://logdetective.com/download"
 MAX_ARCHIVE_AGE_DAYS = 90
 LOG_DIR = os.getenv("LOG_DIR", "/var/log/logdetective")
 LOG_FILE = os.path.join(LOG_DIR, "user_contributions_update.log")
+THREAD_COUNT = int(os.getenv("USER_CONTRIBUTION_UPDATE_THREADS", "4"))
 
 
 # pylint: disable=too-many-locals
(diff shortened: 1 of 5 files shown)


logdetective/server/user_contributions_update.py
```diff
@@ -74,6 +74,7 @@
 MAX_ARCHIVE_AGE_DAYS = 90
 LOG_DIR = os.getenv("LOG_DIR", "/var/log/logdetective")
 LOG_FILE = os.path.join(LOG_DIR, "user_contributions_update.log")
+THREAD_COUNT = 4
```
jpodivin: Please use an env var, so we can parametrize this.

Jany26:
@coderabbitai review

jpodivin (changes requested):

logdetective/server/user_contributions_update.py
Jany26: added `USER_CONTRIBUTION_UPDATE_THREADS` passed via compose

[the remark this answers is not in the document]
Jany26: handled separately via environment variable

jpodivin (approved):

PR #585, opened 2026-08-19
Escape inputs in HTML Jinja templates

Jany26:
Regarding HTML injection into LLM outputs for gitlab jobs. Feels like it's actually impossible to happen, but we can cross out another item from THREAT_MODELS.

EDIT: Adding pre and post fix files (one with valid fields, one with attempts to inject stuff) - that's the `gitlab_full_comment.md.j2` template
- [old-render-okay.html](https://github.com/user-attachments/files/31258513/old-render-okay.html)
- [old-render-error.html](https://github.com/user-attachments/files/31258510/old-render-error.html)
- [render-okay.html](https://github.com/user-attachments/files/31258516/render-okay.html)
- [render-error.html](https://github.com/user-attachments/files/31258514/render-error.html)

## Summary by CodeRabbit

* **Security**
  * Added HTML escaping to generated GitLab comments and responses.
  * Escaped project details, explanations, package names, log and artifact links, and snippet content before display.
  * Updated the threat model to reflect the improved handling of rendered content.


jpodivin (approved):
Have you tested rendering of templates?

PR #588, opened 2026-08-24
Chunk-processing of files from gitlab archives

Jany26:
Added a chunked processing of all accessed files with enforced limit + a small refactor.
Since there is a hard-cap on each archive member, the ratio check can be completely dismissed.

The threat is mitigated only partially, full mitigation assumes a scenario where zip bomb would be the smallest concern.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved GitLab build-log processing to safely handle compressed artifacts.
  - Oversized decompressed logs are now rejected, even when the compressed download itself is within limits.
  - Retrieved logs are processed as decoded text for more consistent downstream display and handling.

- **Documentation**
  - Updated security documentation to reflect the current treatment of compressed build-log artifacts and related risks.


logdetective/server/gitlab.py
```diff
@@ -226,46 +244,48 @@ async def retrieve_and_preprocess_koji_logs(
...
+                # `see <log> for more information`.
+                # Note: it may sometimes say
+                # `see build.log or root.log for more information`, but in
+                # that situation, we only want to handle build.log (for now),
+                # which means accepting only the first match for the regular
+                # expression.
+                contents = read_zip_entry_chunked(
+                    artifacts_zip, zipinfo.filename, gitlab_cfg.max_artifact_size - size_sum
+                ).decode("utf-8")
+                size_sum += len(contents)
```
jpodivin, line 288: This will mean that `size_sum` is counting decoded unicode chars, not bytes.

jpodivin (changes requested):

logdetective/server/gitlab.py
Jany26: right, fixed.

jpodivin (approved):

PR #595, opened 2026-09-08
Create unified task queue for koji analysis requests

jpodivin:
This PR refactors current task handling for Koji analysis requests, and prepares Log Detective for transition to fully asynchronous API. The original `logdetective.database.models.koji` has been replaced by `logdetective.database.models.tasks`. The `TaskAnalysis` table will eventually serve as queue for all tasks submitted to the API. 

To keep database objects fully in memory after committing them, I'm setting `expire_on_commit=False`. This won't impact the operations, as `TaskAnalysis` objects returned  won't be modified after retrieval.

The `AnalyzeRequestMetrics` table now only stores metrics, not tasks or responses. This way we have a clear separation of two concerns. Previously it worked also as a queue. 

While working on this, I rewrote the way we build revisions with alembic so that we don't need to start the entire stack, just the server and database. This makes the process a bit faster and slimmer.


server/config.yml
```diff
@@ -86,7 +86,7 @@ general:
   # generate_solution: true
   # Enable lazy fetching of build artifacts
   # delay_artifact_download: true
-  agent_timeout: 600  # Timeout on agent execution
+  analysis_timeoutt: 600  # Timeout on agent execution
```
Jany26: typo

logdetective/models.py
```diff
@@ -423,7 +423,7 @@ class GeneralConfig(BaseModel):
     generate_solution: bool = True
     delay_artifact_download: bool = False
     # Timeout for execution of analysis in seconds
-    agent_timeout: int = 600
+    analysis_timeout: int = 900
```
Jany26, line 426: I wonder if gunicorn worker timeout accounts for this...

logdetective/database/models/exceptions.py
```diff
@@ -1,16 +1,20 @@
...
 
 
-class KojiTaskAnalysisTimeoutError(Exception):
-    """Exception raised when a koji task analysis has timed out"""
+class TaskAnalysisTimeoutError(Exception):
+    """Exception raised when a task analysis has timed out"""
+
+
+class TaskAlreadyRetrievedError(Exception):
+    """Exception raised when task was already retrieved"""
```
Jany26: this exception is not used, maybe some future plans?

Makefile
```diff
@@ -17,32 +17,45 @@ server-up:
...
+		-v $(PWD)/logdetective:/src/logdetective:ro,z \
 		-v $(PWD)/alembic:/src/alembic:rw,z \
 		-v $(PWD)/alembic.ini:/src/alembic.ini:ro,z \
 		--network logdetective_default \
 		localhost/logdetective/server:latest \
 		bash -c "cd /src && python -m alembic revision -m \"$(CHANGE)\" --autogenerate"
 
-	@echo "WARNING: Shutting down server..."
-	$(COMPOSE_ENGINE) down server
+	podman rm logdetective_postgres_1 --force
```
Jany26: `$(CONTAINER_ENGINE)`

logdetective/database/models/tasks.py
```diff
@@ -0,0 +1,263 @@
...
+
+    @classmethod
+    @retry_database_error
+    async def get_task_by_id(cls, task_id: uuid_type) -> TaskAnalysis:
+        """Get analysis task response by task id"""
+
+        query = (
+            select(cls).filter(cls.task_id == task_id).with_for_update(skip_locked=True)
+        )
+        return await cls.get_task(query=query, identifier=str(task_id))
```
Jany26, line 251: Just noting that this is also unused, but i guess it might be useful later.
Also it might be worth to note where the external ID will come from?

logdetective/database/models/tasks.py
```diff
@@ -0,0 +1,263 @@
...
+                        f"Task {identifier} analysis has timed out"
+                    )
+
+                # Task analysis is still in progress, so we need to let the
+                # consumer know
+                raise TaskNotAnalyzedError(
+                    f"Task {identifier} analysis is still in progress"
+                )
+
+            task_analysis.response_returned_at = datetime.now(timezone.utc)
```
Jany26, line 237: is it desirable to overwrite the timestamp on every retrieval?

logdetective/database/models/exceptions.py
Jany26: this exception is not used, maybe some future plans?

Jany26 (changes requested):
Couple of small things. Thanks for reworking the make target.

logdetective/database/models/exceptions.py
jpodivin: Good catch, this is from earlier prototype. Turns out it was a bad idea.

logdetective/database/models/tasks.py
jpodivin: This is for general analysis tasks. For Koji, we have an external ID supplied based on Koji job.

logdetective/database/models/tasks.py
jpodivin: I don't really expect the response to be retrieved more than once, but having a record of when it was accessed is necessary. Eventually we may start pruning entries that were accessed long time ago. Having two columns, one for first access and another for last, is possible. But I don't see any benefit it.

logdetective/models.py
jpodivin: After we switch `/analysis` to queue based processing, the gunicorn timeout will be decoupled from the agent timeout. This PR brings us into sort of a transition state.

Jany26 (approved):

PR #596, opened 2026-09-10
Multiple tokens for authentication

jpodivin:
Instead of using only one token, we will allow for mutiple tokens, one for each user. The  `api_tokens.yml` will be checked for duplicates. Since those would indicate improper credential management, the server will reject this configuration and raise exception.

In case when path to the token file is unset, the server will not require any authentication for API calls.

Metrics will now contain information about caller who requested analysis, based on the token used.


logdetective/utils.py
```diff
@@ -33,6 +37,27 @@
 )
 
 
+def load_api_tokens(path: str | None) -> APITokens | None:
+    """Load and validate named API bearer tokens from a YAML file."""
+    if not path:
+        LOG.warning(
+            "Log Detective launched without set API tokens. "
+            "All request will be accepted without authentication!"
```
Jany26: requests - plural

Jany26 (changes requested):
Aside from the small typo, looks okay.

Also for clarification, for metrics:
- If we provide token name, we only obtain metrics for data submitted with that token.
- If we don't provide token, we get all request/response data.
- Metric data (if api-tokens autorization is on) is available only if some valid token is used.

jpodivin:
> Metric data (if api-tokens autorization is on) is available only if some valid token is used.

Pretty much. We've gated metrics on tokens before. But now you can measure usage on user basis.

Jany26 (approved):

PR #597, opened 2026-09-11
Add hostname to installed packages

jpodivin:
During compose startup, two identical messages about missing `hostname` binary show in log.

```
sh: line 1: hostname: command not found
```

Investigation has revealed that they originate from Alembic migration, and server start up, and both involve `blkid` calls in parent process. The working hypothesis, is that this is a form of telemetry fingerprinting, most probably coming from Sentry. It should be noted, that this error message does not appear to indicate any issue with our application. As of now, there are no bugs associated with it.

That being said. I would prefer not to have a silently broken in our server. 

I've used this script, mounted to `/bin/hostname` to track down PID of the caller:

```
#!/bin/sh
echo "--- $(date) ---" >> /tmp/hostname_trap.log

CURRENT_PID=$PPID

while [ "$CURRENT_PID" -gt 1 ] && [ -d "/proc/$CURRENT_PID" ]; do
    CMD=$(tr '\0' ' ' < /proc/$CURRENT_PID/cmdline)
    echo "PID $CURRENT_PID: $CMD" >> /tmp/hostname_trap.log
    read pid comm state ppid rest < /proc/$CURRENT_PID/stat
    CURRENT_PID=$ppid
done
echo "" >> /tmp/hostname_trap.log
```

## Summary by CodeRabbit

- **Chores**
  - Added the `hostname` utility to the container environment.


Jany26 (approved):

Jany26 (commented):
Also, are you sure we want to merge it against `main`?

PR #599, opened 2026-09-17
Switching enum values to upper case to simplify postgresql integration

jpodivin:


Jany26:
what about ~~EndpointType, TimePeriod, and~~ Forge values? Or that would be complicated since they already are in the DB?

Jany26 (approved):

PR #601, opened 2026-09-18
Async API 

jpodivin:
- Application records and jobs are processed atomically
- Durable polling and cancellation with idempotency, data retention, and reconciliation
- Generic, Koji, and GitLab analysis processes are moved into asynchronous workers
- Two new containers, for workers and for migrations

Presence non-empty legacy task data in database blocks migration, to ensure we have a clean slate.

BREAKING CHANGE: Since analysis endpoints return `202 Accepted` task envelopes. All clients must be updated to recognize the new contract and use poll `GET /tasks/{id}` to get results.

For testing of the new API, use `scripts/test_request.sh` with chosen Koji results.


logdetective/database/models/tasks.py
```diff
@@ -12,252 +22,751 @@
...
+
+        Returns:
+            ``True`` when the fenced update published the result; ``False`` when the
+            task changed state or ownership before publication.
+
+        Raises:
+            RuntimeError: If the related request metrics record disappeared.
+        """
+        from logdetective.compressors import (  # pylint: disable=import-outside-toplevel
+            LLMResponseCompressor,
```
Jany26: I think this import can be moved to the top of the module.

logdetective/gitlab.py
```diff
@@ -227,97 +289,14 @@ async def retrieve_and_preprocess_koji_logs(
...
-    log_url = f"{gitlab_cfg.url}/{gitlab_cfg.api_path}/projects/{job.project_id}/jobs/{job.id}/artifacts/{log_path}"  # pylint: disable=line-too-long
-    LOG.debug("Returning contents of %s%s", gitlab_cfg.url, log_url)
-
-    # Return the log as a file-like object with .read() function
-    return log_url, artifacts_zip.open(log_path)
+    try:
+        await run_blocking(job.artifacts, streamed=True, action=tempfile.write)
+        return await run_blocking(_select_koji_log, gitlab_cfg, job, tempfile)
+    finally:
+        await run_blocking(tempfile.close)
```
Jany26: This is a little weird, I am sure that I wrapped the tempfile handling in a context manager. I guess `ld_v5` was branched before that change was in main. Anyway this will careful conflict resolution during rebase to main (especially since a lot of functionality was moved elsewhere), hopefully the changes account for those context manager changes.

logdetective/gitlab.py
```diff
@@ -50,6 +52,64 @@
 FAILURE_LOG_REGEX = re.compile(r"(\w*\.log)")
 
 
+def _select_koji_log(gitlab_cfg, job, tempfile):
```
Jany26: during rebase to main, make sure that this function reflects changes from #588

logdetective/server.py
```diff
@@ -240,374 +132,228 @@ def authenticate_api_token(
...
 
-async def get_artifacts_from_payload(
-    payload: AnalysisRequest,
-    http_session: aiohttp.ClientSession,
-    request_size: int,
-) -> dict[str, str | RemoteLog]:
-    """Retrieve artifact contents based on the type of artifact.
-    Raise ValueError on unsupported element types."""
-    build_artifacts: dict[str, str | RemoteLog] = {}
+async def validate_request_size(request: Request) -> int:
```
Jany26: original function was non-async, I don't think the changes warrant the switch to async.

logdetective/server.py
```diff
@@ -240,374 +132,228 @@ def authenticate_api_token(
...
+        if task.response is None:
+            raise RuntimeError("Completed task has no persisted response")
+        response = LLMResponseCompressor.unzip(task.response)
+        if task.task_type == TaskType.KOJI:
+            if task.task_metadata is None:
+                raise RuntimeError("Completed Koji task has no metadata")
+            external_id = task.task_metadata.get("task_id")
+            log_file_name = task.task_metadata.get("log_file_name")
+            if external_id is None or log_file_name is None:
+                raise RuntimeError("Completed Koji task metadata is incomplete")
```
Jany26: is it worth to consider making `test_metadata` a pydantic model too?

logdetective/models.py
```diff
@@ -292,6 +310,27 @@ class KojiResponse(BaseModel):
...
+    code: str
+    message: str
+
+
+class TaskResponse(BaseModel):
+    """Stable response envelope used from admission through termination."""
+
+    model_config = ConfigDict(populate_by_name=True)
+    id: UUID
+    task_type: Literal["generic", "koji"] = Field(alias="taskType")
```
Jany26: Codex thinks this should be in-line with `TaskType` or `EndpointType` enum. If not, maybe a note should be here explaining why. I think it's worth adding a note anyway.

logdetective/task_execution.py
```diff
@@ -0,0 +1,235 @@
...
+        metadata={"log_file_name": filename, "task_id": payload.task_id},
+    )
+
+
+async def _execute_gitlab(
+    payload_data: dict[str, Any], metrics_id: int | None
+) -> TaskOutcome:
+    forge = Forge(payload_data["forge"])
+    job_hook = JobHook.model_validate(payload_data["job_hook"])
+    config = SERVER_CONFIG.gitlab.instances[forge.value]
```
Jany26: Codex thinks there's a possible unhandled KeyError here (if API and worker configs diverge). I'm not sure but I think this is unlikely.

logdetective/task_execution.py
```diff
@@ -0,0 +1,235 @@
...
+                http,
+                forge,
+                job_hook,
+                get_chat_model(SERVER_CONFIG.inference),
+                payload_data.get("api_token_name"),
+                metrics_id,
+            )
+    finally:
+        await run_blocking(connection.session.close)
+    return TaskOutcome(response=response, metadata={})
```
Jany26: Codex: Return empty task outcome for webhooks to avoid retaining payloads - `_execute_gitlab` returns full analysis payloads that are persisted by `publish_result`, even though `task_representation` explicitly hides GitLab task results from clients. Returning `TaskOutcome(response=None, metadata={})` calls `complete_without_result` to prevent storing unexposed data. This is a data retention and storage optimization issue.

logdetective/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Returns:
+        ``None`` after webhook processing is durably marked complete.
+    """
+    await _execute(context, task_id)
+
+
+@app.periodic(cron="*/1 * * * *", periodic_id="reconcile-analysis-jobs")
+@app.task(queue="maintenance", queueing_lock="reconcile-analysis-jobs")
+async def reconcile_analysis_jobs(timestamp: int) -> None:
```
Jany26: we could use more logging here, especially for cancellation and stalling

logdetective/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+        except NoResult:
+            await TaskAnalysis.mark_worker_lost(
+                task.procrastinate_job_id,
+                SERVER_CONFIG.task_queue.retention_days,
+            )
+            continue
+        if status in (Status.FAILED, Status.SUCCEEDED, Status.ABORTED, Status.CANCELLED):
+            if task.state == AnalysisState.CANCELLING and status in (
+                Status.ABORTED,
+                Status.CANCELLED,
```
Jany26: is marking a cancelled failed/succeeded job with `worker_lost` correct?

logdetective/task_execution.py
```diff
@@ -0,0 +1,235 @@
...
+                payload.task_id,
+                max_size=SERVER_CONFIG.koji.max_artifact_size,
+            )
+    finally:
+        await _close_koji_session(session)
+
+    response = await analyze_artifacts(
+        {filename: await run_blocking(sanitize_artifact, content)},
+        get_chat_model(SERVER_CONFIG.inference),
+    )
```
Jany26: It looks like there is a discrepancy here. We configure `koji.analysis_timeout`, yet here the analyze_artifacts is outside of the scope with the timeout. Shouldn't the `koji.analysis_timeout` should override the default `api_timeout` or `analysis_timeout`?

logdetective/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+                await TaskAnalysis.confirm_cancelled(
+                    task.task_id,
+                    task.procrastinate_job_id,
+                    SERVER_CONFIG.task_queue.retention_days,
+                )
+            else:
+                await TaskAnalysis.mark_worker_lost(
+                    task.procrastinate_job_id,
+                    SERVER_CONFIG.task_queue.retention_days,
+                )
```
Jany26: is marking a cancelled failed/succeeded job with `worker_lost` correct?

logdetective/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+        try:
+            status = await app.job_manager.get_job_status_async(
+                task.procrastinate_job_id
+            )
+        except NoResult:
+            await TaskAnalysis.mark_worker_lost(
+                task.procrastinate_job_id,
+                SERVER_CONFIG.task_queue.retention_days,
+            )
```
Jany26: Codex: A task in `CANCELLING` state whose job is missing yields `NoResult` and will eventually be mapped to `ERROR`. 

When `cancel_job_by_id_async` returns `False` in `request_cancellation` (job already gone from `Procrastinate`), the task stays `CANCELLING`. On the next reconciliation tick, `get_job_status_async` raises `NoResult`. The task transitions to `ERROR` with `error_code="worker_lost"`. A task whose cancellation was explicitly requested should reach `CANCELLED`, not `ERROR`. The reconciler's `NoResult` branch should check whether the task is in `CANCELLING` state and call `confirm_cancelled` instead.

logdetective/worker.py
```diff
@@ -0,0 +1,25 @@
...
+        operation, the call blocks while jobs are consumed.
+    """
+    app.run_worker(
+        queues=["analysis", "maintenance"],
+        concurrency=SERVER_CONFIG.task_queue.concurrency,
+        shutdown_graceful_timeout=(
+            SERVER_CONFIG.task_queue.shutdown_graceful_timeout
+        ),
+        stalled_worker_timeout=SERVER_CONFIG.task_queue.stalled_worker_timeout,
+    )
```
Jany26: won't long analyses starve maintenance? if it's important that maintenance runs like clockwork, we should have a dedicated worker for it.

Jany26 (changes requested):
Some of Codex's findings seem reasonable (especially the one with `reconcile_analysis_jobs`). Other than that I noticed a couple of cosmetic oversights and stuff that we will need to keep an eye on when we rebase `ld_v5` on `main`.

logdetective/gitlab.py
jpodivin: Yes. Merging ld_v5 back into main will be a complicated affair.

logdetective/database/models/tasks.py
jpodivin: Yes. There is no reason for it to be here.

logdetective/server.py
jpodivin: They do not. Back sync it is.

logdetective/models.py
jpodivin: Good idea.

logdetective/task_execution.py
jpodivin: It would mean that we stored malformed or incorrect dictionary in the database.

logdetective/server.py
jpodivin: Yes.

logdetective/tasks.py
jpodivin: Yes.

logdetective/task_execution.py
jpodivin: Setting `response=None` is probably the better option here. The consistency is still worth something if we ever decide that we want to work with webhooks in a different way.

logdetective/task_execution.py
Jany26: yeah, this finding was not useful

logdetective/task_execution.py
Jany26: I see the config option was changed to `koji.api_timeout` and `koji.retrieval_timeout`. Not sure why GitHub didn't pick up on this change.

logdetective/database/models/tasks.py
```diff
@@ -12,252 +21,777 @@
...
+            return
+        metric = await session.get(AnalyzeRequestMetrics, metric_id)
+        if metric is None:
+            raise RuntimeError("Analysis metric disappeared before task completion")
+        metric.analysis_completed_at = finished_at
+        if response_length is not None:
+            metric.response_length = response_length
+
+    @classmethod
+    async def admit(  # pylint: disable=too-many-arguments,too-many-locals,too-many-branches
```
Jany26: Codex: Expired task IDs return an unusable accepted operation

`admit()` returns a matching record without checking `expires_at`, while `get_owned()` hides expired records. During the interval before the hourly expiry job deletes it, an identical retry receives `202` and a `Location` that immediately returns `404`. Reject expired IDs with `409`, or delete/recreate the expired record atomically before admission.

logdetective/gitlab.py
Jany26: Codex: GitLab webhook response time is overwritten by worker completion

Admission records `response_sent_at` when the webhook can return `204`, but `update_metrics()` replaces it after inference and GitLab comments finish. This makes `average_response_time` report worker duration rather than webhook acknowledgement latency. Preserve the admission timestamp; update only response length/completion metadata for an admitted GitLab task.

Jany26 (changes requested):
- 2 findings from codex
- we should regenerate mermaid db diagrams
- seems that 2 comments from the previous review were overlooked, or are non-actionable and a comment explaining that is missing

logdetective/tasks.py
jpodivin: Yes. Active task should not have job status `Status.SUCCEEDED`. If it has, it indicates a queue failure. Name of the method isn't exactly well chosen. I'll change it.

logdetective/gitlab.py
jpodivin: Yes, this can actually happen.

logdetective/worker.py
jpodivin: I've split workers in two containers and added concurrency to the analysis worker. This should keep the operation smooth.

logdetective/database/models/tasks.py
jpodivin: Correct, this is an inconsistency.

logdetective/gitlab.py
jpodivin: We'll have to check a lot of things manually. Hopefully we can rely on tests to catch at least some of the issues that can show up.

Jany26 (approved):

